PR #1, opened 2013-01-14
Support x86

NicolasT:
This branch tracks support for the x86 platform for the 1.5 branch.

Don't merge this pull request, this should be done in Mercurial, if accepted.


NicolasT:
Notice TravisCI builds fail (for now) since their success depends on #2.

NicolasT:
No longer using TravisCI.

PR #2, opened 2013-01-14
Add support files for TravisCI

NicolasT:
This branch adds support files for TravisCI (http://travis-ci.org), see https://travis-ci.org/Incubaid/arakoon.


NicolasT:
Note builds on TravisCI builds will fail (for now) since their success depends on #1 (TravisCI hosts are x86).

NicolasT:
No longer using TravisCI.

PR #6, opened 2013-02-13
ARAKOON-360: Allow multiple 'preferred master' nodes

NicolasT:


toolslive:
Where is the (unit|system) test that 
- the way we configure things with the qshell extension is not broken.
- shows Arakoon behaves as it should?

NicolasT:
Added tests. Also ran some more manual validation of failover behavior, seems to work as desired/expected.

PR #9, opened 2013-02-20
Nursery

toolslive:
Functionality passes the 1.x nursery tests.


src/client/remote_admin_test.ml
```diff
@@ -113,11 +194,5 @@ let set_route_delta port () =
...
   __wrap__ port conversation
 
-let suite = 
-  let w f = Extra.lwt_bracket setup f teardown in
-  "nursery" >:::
-    ["set_interval" >:: w (set_interval 6666);
-     "get_fringe"  >:: w (get_fringe  5555);
-     "set_routing"  >:: w (set_route_delta  4444);
-    ]
+*)
```
NicolasT, line 197: Should commented code be kept around?

src/hope/bstore.ml
```diff
@@ -85,19 +85,22 @@ module BStore = (struct
         let rec _inner (tx: BS.tx) = function
           | Core.SET (k,v) ->
               begin
-                Lwtc.log "set: %s" k >>= fun () -> 
+                (* Lwtc.log "set: %s" k >>= fun () ->
```
NicolasT, line 88: Remove commented code (also below)?

OK to remove _check_interval?

src/hope/bstore.ml
```diff
@@ -311,17 +334,92 @@ module BStore = (struct
...
+                  let start' = Some (last_key kvs_rev) in
+                  let mem' = List.fold_left 
+                    (fun mem (k,v) ->
+                      mem + String.length k + String.length v) mem kvs_rev 
+                  in
+                  if mem' < limit
+                  then loop start' mem' acc'
+                  else Lwt.return (acc', mem')
+              in
+              loop (Some "\xff") 0 []
```
NicolasT, line 390: Is starting at 0xff 'far enough to the right'?

src/hope/bstore.ml
toolslive, line 88: the check interval indeed should be introduced, but only after we have something like a 'management sequence'.
so the introduction earlier was a bit premature.

src/hope/bstore.ml
toolslive, line 390: nope. but for that I need to adapt baardskeerder.

src/client/remote_admin_test.ml
toolslive, line 197: These were tests from the 1.x era, and the suite was not part of the 2.x.  
So I re-enabled the suite and made the test I needed succeed again. If we encounter a bug in the other parts of the protocol, these could be used as guidance.

NicolasT:
LGTM, but might be worth to put some of the dev TODO items in JIRA so they don't get forgotten?

PR #11, opened 2013-02-25
Arakoon 347

blackmath86:
# This branch till now ignores:
- The connecting dss command line utility also reports its version/revision information so that compatibility problems can be confirmed if necessary.
- DSS uses different unique ids for “manager” threads within the daemon process, so that the log lines they emit can also be correlated.
- Log fd, peer address
# Some logging sections were added
- paxos
- client_protocol
- tcp_messaging

The default behavior for the logger will be determied in server.ml#76, its set to debug as default.
It's user choice to set env_variable LWT_DEBUG as stated in README. 
# These features were added to the current branch"
- Associate a unique connection id with each connection
- Use connection id as log prefix when working with it (see Lwt.new_key, Lwt.with_value
  and Lwt.get to propagate the id through the lwt code using threadlocal
  storage)
- Log fd, peer address and port and local address and port at connect, accept and at shutdown / disconnect, with the connection id as prefix
- expose the Arakoon decision process on consensus and election. Paxos
- Measure every write to disk and log the time it took if this writeoperation
  takes more than 1 second
- When rebuilding a tc file from the tlf files, it's usefull to log what tlf file is being replayed
# The "milli-second(-like) " option is implemented in 1.5.0 where the current branch is based on.

### DIFF
--- a/src/msg/tcp_messaging.ml
+++ b/src/msg/tcp_messaging.ml
@@ -28,7 +28,7 @@ open Lwt_buffer
 open Network
 
 type connection = Lwt_io.input_channel * Lwt_io.output_channel
-
+let section = Lwt_log.Section.make "tcp_messaging"
 
 let never () = Lwt.return false 
 let no_callback () = Lwt.return ()
@@ -92,7 +92,7 @@ object(self : # messaging )
 	
 	
   method private __die__ () = 
-    Lwt_log.debug_f "tcp_messaging %s: cancelling my threads" me >>= fun () ->
+    Lwt_log.debug_f ~section "tcp_messaging %s: cancelling my threads" me >>= fun () ->
     List.iter (fun w -> Lwt.cancel w) _my_threads;
     _my_threads <- [];
     Lwt.return () 
@@ -110,6 +110,7 @@ object(self : # messaging )
 	
   method send_message m ~source ~target =
     let tq = self # _get_send_q ~target in
+    Lwt_log.debug_f ~section "MSG: %s => sending msg(%s) to => %s" source (Message.string_of m) target >>= fun () ->
     Lwt_buffer.add (source, target, m) tq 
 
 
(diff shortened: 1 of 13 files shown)


NicolasT:
This looks fishy. Why does the PR contain much older commits which should be on 1.6 already, but having different commit IDs here? This looks like a rebase-gone-wrong.

src/msg/tcp_messaging.ml
```diff
@@ -128,13 +129,18 @@ object(self : # messaging )
 
   method recv_message ~target =
     let q = self # get_buffer target in 
-    Lwt_buffer.take q
+    Lwt_buffer.take q >>= fun msg ->
+    let a,b = msg in
+    Lwt_log.debug_f ~section "MSG: receiving msg(%s) from %s" (Message.string_of a) b >>= fun () ->
```
toolslive: I don't think you want to do this:
Message.string_of builds a string that includes the payload.

src/node/local_store.ml
```diff
@@ -331,8 +331,17 @@ let _set_master bdb master (lease_start:int64) =
 let _with_tx (db: Camltc.Hotc.t) (f:B.bdb -> 'a Lwt.t) =
   Camltc.Hotc.transaction db
 	(fun db ->
+      let t0 = Unix.gettimeofday() in
 	  f db >>= fun (a:'a) ->
-	  Lwt.return a)
+      let t = ( Unix.gettimeofday() -. t0) in
+      if t > 1.0 
+      then begin
+        Lwt_log.info_f "Operation lasted for %fs" t >>= fun () ->
```
toolslive: logging statement is too fuzzy. 
something like "Tokyo cabinet transaction took %fs" would be better.

src/msg/tcp_messaging.ml
blackmath86: kind will be enough?

PR #12, opened 2013-03-04
Jenkins: Use LXC containers

NicolasT:
These commits rework the Jenkins/CI setup to use LXC containers (see http://confluence.incubaid.com/display/ENG/Setup+of+cislave11 for more information about the setup).

Some tests might still fail (due to missing permissions since no longer running as root?), but it would be useful to merge the current changes as-is so the 1.6 jobs can be put in place.


NicolasT:
Hmh, wrong target branch :-(

PR #14, opened 2013-03-04
Arakoon 364

toolslive:


NicolasT:
LGTM, although it occurred to me you use List.rev in https://github.com/Incubaid/arakoon/pull/14/files#L7R88 whilst I thought lists are encoded on-the-wire in reverse order, to be decode'able without reversing?

toolslive:
yes, lists are encoded on the wire in reverse order, to be decodable without reversing.
but a sequence does not seem to be a list but an array,
we could change this but then we need to change all clients.

PR #16, opened 2013-03-06
Arakoon 326

toolslive:


extension/test/server/quick/system_tests_basic.py
```diff
@@ -394,7 +418,7 @@ def test_3_nodes_stop_all_start_slaves ():
     cli.set(key,value)
     master = cli.whoMaster()
     slaves = filter( lambda node: node != master, C.node_names )
-    
+    logging.debug("slaves=%s", slaves)
```
NicolasT, line 429: This will be fixed through #15

NicolasT:
Is there a specific exit code being used upon detection of a missing 'closed' marker, or does this depend on the work in ARAKOON-358?

NicolasT:
Might be useful to have a test which validates the 'missing close marker' is reported (~ exit code) after reporting corrupt tlog or corrupt TC.

NicolasT:
LGTM, if 'exit reason order' is handled correctly.

PR #18, opened 2013-03-11
Fix 326

toolslive:


NicolasT:
Looks like some tests might fail: http://cimaster.incubaid.com/job/arakoon-git-base/71/testReport/?

'left' and 'right' still running.

PR #19, opened 2013-03-13
Use separate exit codes for distinct error conditions

NicolasT:
This allows for wrapper scripts to take action upon node failure.

See: ARAKOON-358
See: http://jira.incubaid.com/browse/ARAKOON-358


toolslive:
Looks fine.

PR #22, opened 2013-03-20
flag that enables logging of every client protocol operation

domsj:


src/client/client_protocol.ml
```diff
@@ -128,58 +128,74 @@ let handle_sequence ~sync ic oc backend =
...
 
-let one_command (ic,oc) (backend:Backend.backend) =
+let one_command (ic,oc) (backend:Backend.backend) log_commands =
+  let get_string = function
+    | None -> ""
+    | Some s -> s
+  and log_command_f =
+    begin
+      if log_commands then
+        Lwt_log.info_f
```
toolslive, line 135: Is this what we want?
I would think, if the flag is there, we log in debug, else we don't log at all.

NicolasT:
This changeset contains lots of hunks which look as if your editor changed tabs to spaces (or the other way around). That's rather unfortunate :-/

src/client/client_protocol.ml
```diff
@@ -128,58 +128,74 @@ let handle_sequence ~sync ic oc backend =
...
-      begin
-        Llio.input_string ic >>= fun client_id ->
-	    Llio.input_string ic >>= fun cluster_id ->
-        backend # hello client_id cluster_id >>= fun (rc,msg) ->
-        response_rc_string oc rc msg
-      end
+        begin
+          Llio.input_string ic >>= fun client_id ->
+	  Llio.input_string ic >>= fun cluster_id ->
+          log_command_f "PING: client_id=%S cluster_id=%S" client_id cluster_id >>= fun _ ->
```
NicolasT: Use 'fun () ->' instead of a wildcard pattern (e.g. to make sure the action isn't partially applied to its arguments).

src/client/client_protocol.ml
```diff
@@ -128,58 +128,74 @@ let handle_sequence ~sync ic oc backend =
 
   end
 
-let one_command (ic,oc) (backend:Backend.backend) =
+let one_command (ic,oc) (backend:Backend.backend) log_commands =
+  let get_string = function
```
NicolasT: Consider using 'None' and 'Some "..."' as return value instead of 'empty string' (which might be a valid 'Some' value).

NicolasT:
Next to the changes to the 'node' core, the Pylabs extension should gain support to configure this setting as well.

src/client/client_protocol.ml
domsj: I explicitly wrote this function to convert a string option to a string so I could pass the result to e.g. Lwt_log.info_f "%S" (in this case passing it log_command_f which I defined a couple of lines lower).
Is there a better way to accomplish this?

src/client/client_protocol.ml
domsj: Hmm did that for the most part, somehow missed this one. Will update.

src/client/client_protocol.ml
NicolasT: I meant something along the lines of

``` ocaml
let get_string = function
  | None -> "None"
  | Some s -> Printf.sprintf "Some %S" s
```

or, if 'maybe' would be defined somewhere in general (OTOH):

``` ocaml
let maybe v f = function
  | None -> v
  | Some v' -> f v'

let get_string = maybe "None" (Printf.sprintf "Some %S")
```

domsj: Ok, I misunderstood.
The change you propose surely makes the difference between None an d Some "" more clear in the logging.
Will have to change the %S to %s then

PR #26, opened 2013-04-03
Cleanup tree

NicolasT:
These commits remove some old and unused files from the tree:
- _ROOT/OCAML/include/event.h_, most likely added by accident.
- _src/client/c/_, the deprecated C client.
- _src/client/php/_ and _jenkins/php/_, moved into @Incubaid/pharakoon
- _hudson/_, containing some old and seemingly unused CI scripts.
- _tools/patch.lwt_io_ since we use upstream Lwt for quite a while.
- _tools/check_and_fix_tlog_8.2_to_0.1x.0.py_, since 0.8 and 0.1x have been unsupported for quite a while.

I will run everything through CI first to make sure no jobs break. This is more of a heads-up.


NicolasT:
_base_ and _quick_ passed (well, shaky tests failed, but not due to changes in these commits).

PR #29, opened 2013-04-06
no output on stdout; mem_tlogcollection improvement

toolslive:
the  change that fixes the tests is mem_tlogcollection get_tlog_from_i.
the rest is just a long overdue clean up action


NicolasT:
LGTM (although we should really look into ocp-indent'ing the code in a single commit :smiley: )

PR #31, opened 2013-04-10
fix compress tlog removing the .tlog file without producing the .tlf

domsj:


NicolasT:
If this passes system tests, LGTM, although I'm not a big fan of the tool unlinking the source file.

PR #32, opened 2013-04-11
fix the addNode isLocal behaviour

domsj:
I'm opinionated.

### DIFF
--- a/extension/server/ArakoonManagement.py
+++ b/extension/server/ArakoonManagement.py
@@ -235,7 +235,7 @@ class ArakoonCluster:
         @param tlogDir :  the directory used for tlogs (if none, home will be used)
         @param isLearner : whether this node is a learner node or not
         @param targets : for a learner node the targets (string list) it learns from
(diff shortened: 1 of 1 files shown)


extension/server/ArakoonManagement.py
```diff
@@ -235,6 +235,7 @@ def addNode(self,
         @param tlogDir :  the directory used for tlogs (if none, home will be used)
         @param isLearner : whether this node is a learner node or not
         @param targets : for a learner node the targets (string list) it learns from
+        @param isLocal : wether this node is a local node and should be added to the local nodes list
```
toolslive: whether.

extension/server/ArakoonManagement.py
```diff
@@ -295,6 +296,9 @@ def addNode(self,
 
         config.write()
 
+        if isLocal:
+            self.addLocalNode(name)
+
```
toolslive, line 301: so it boils down to a shortcut. and because the default is False, the is-present check in addLocalNode will not give problems.
I think this is perfectly acceptable.

PR #33, opened 2013-04-11
Arakoon 347 2

domsj:
This concerns the extra logging.
Added the following sections: client_protocol, paxos & tcp_messaging.

I also took a look at https://github.com/Incubaid/arakoon/commit/4a89d1f01de3a5eb246502a9a1a57ab372de1b75 (347 from Moataz).
Main things I don't have, judged from a quick scan over his commit would be:
- I have not added the logging to mem_store.ml;
- I did not remove the log_o logging in sync_backend.ml
- I have not explicitly logged n and i in the paxos related files. That information should be mostly available from the incoming and outgoing messages.

Passes CI tests.

### DIFF
--- a/cfg/arakoon.ini
+++ b/cfg/arakoon.ini
@@ -38,6 +38,7 @@ cluster_id = ricky
 # available levels are: debug info notice warning error fatal
 client_protocol = debug
 paxos = debug
+tcp_messaging = debug
 
 [arakoon_0]
 #comma separated list of all ip addresses used to identify that node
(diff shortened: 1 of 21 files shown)


NicolasT:
log_commands seems to be removed. What's the related section name?

domsj:
@NicolasT related section name is client_protocol, see https://github.com/Incubaid/arakoon/blob/ARAKOON-347-2/src/client/client_protocol.ml

PR #34, opened 2013-04-18
sync tlog & database on close & incoming .tlfs in .tmp file

toolslive:


src/tlog/tlc2.ml
```diff
@@ -645,8 +646,10 @@ object(self: # tlog_collection)
   method save_tlog_file name length ic =
     (* what with rotation, open streams, ...*)
     let canon = Filename.concat tlog_dir name in
-    Lwt_log.debug_f "save_tlog_file: %s" canon >>= fun () ->
-    Lwt_io.with_file ~mode:Lwt_io.output canon (fun oc -> Llio.copy_stream ~length ~ic ~oc)
+    let tmp = canon ^ ".tmp" in
+    Lwt_log.debug_f "save_tlog_file: %s" tmp >>= fun () ->
+    Lwt_io.with_file ~mode:Lwt_io.output tmp (fun oc -> Llio.copy_stream ~length ~ic ~oc) >>= fun () ->
```
NicolasT, line 651: For robustness this should somehow use `O_CREAT | O_EXCL` to be on the safe side, and maybe `unlink` the _.tmp_ file first (with a catch in case it doesn't exist).

NicolasT, line 651: Hmh, even that's not entirely safe... But at least somewhat safer :smiley:

NicolasT:
Maybe squash the camltc patches?

src/node/local_store.ml
```diff
@@ -586,6 +586,7 @@ object(self: #store)
 
   method close () =
     _closed <- true;
+    Camltc.Hotc.sync db >>= fun () ->
```
toolslive: doesn't really work:
"invalid operation"
from tokyo cabinet in 
universe:0:collapser:0:collapser_test:0:collapse_many

domsj:
is the manual compression of tlog files done through a temp file? (probably not so important, but while you're at it ...)

PR #37, opened 2013-04-23
discover other master for certain Accepts

toolslive:


src/paxos/multi_paxos_fsm.ml
```diff
@@ -609,30 +609,40 @@ let wait_for_accepteds constants state (event:paxos_event) =
...
                 begin
                   log_f me "wait_for_accepteds: dropping accept with n = %s and i = %s" 
                     (Sn.string_of n) (Sn.string_of i') >>= fun () ->
                   Fsm.return (Wait_for_accepteds state)
                 end
-            | Accept (n',i',v') -> (* n' = n *)
-	            begin
-                  log_f me "wait_for_accepteds: got accept with n = %s and higher i = %s" 
-                    (Sn.string_of n) (Sn.string_of i') >>= fun () ->
+	        | Accept (n',i',v') (* n' >= n || i' > i *)->
```
domsj: shouldn't that comment say n' >= n && i' > i according to the code above?

toolslive: yes, that stronger assertion is also true.

PR #40, opened 2013-04-26
cmd line option to replay tlog

toolslive:


src/main/main.ml
```diff
@@ -241,6 +240,8 @@ let main () =
   and left_cluster = ref ""
   and separator = ref ""
   and right_cluster = ref ""
+  and tlog_dir = ref "."
```
NicolasT: If I understand correctly, the tlog folder is a non-optional argument. As such, maybe don't use "." as initialization value but the empty string instead? This might be confusing to readers.

NicolasT:
LGTM

domsj:
Ok

PR #42, opened 2013-04-29
ARAKOON-379: dropMaster hangs on 1 node cluster

toolslive:
added check & system test


domsj:
looks good

NicolasT:
I'm not sure overloading the 'wrong cluster' error is a good idea (although I seem not to find the place where it's currently used?) The client requesting the operation is connected to the cluster it intends to (otherwise the handshake failed).

extension/test/server/quick/system_tests_basic.py
```diff
@@ -301,6 +301,22 @@ def tes_and_set_scenario( start_suffix ): #tes is deliberate
...
+    client = C.get_client()
+    master = client.whoMaster()
+    cluster = C._getCluster()
+    #assert_raises(Exception, cluster.dropMaster, master) # want to see assert rc too
+    try:
+        cluster.dropMaster(master)
+        raise Error
+    except Exception, e:
+        rc = e.args[0]
+        assert_equals (rc,6, "wrong rc %i" % rc)
```
NicolasT: You could use `ArakoonProtocol.ARA_ERR_WRONG_CLUSTER` (or something alike) instead of a magic number.

PR #44, opened 2013-04-30
Transaction changes

domsj:
Contrary to it's name this branch contains the following changes:
- added sub i tracking (represented by the 'j' counter) to the store, to ensure transactional application of paxos values (with multiple updates) to the store, should the node die while applying these updates
- added with_transaction to the store. the supplied transaction object is as off now required to perform updates to the store.
- added a mechanism (with_transaction_lock) to lock the aquirement of a transaction to help in ensuring the application of a paxos value to the store happens transactionally
- batched multiple tlog entries into 1 store-transaction when performing catchup
- tests for all of the above

suggestions, especially related to my transaction_lock thingie are appreciated.

Other possible changes:
- make batch size configurable
- add batching for the updates inside 1 paxos value (in regular non catchup regime)

I still need to run the integration tests, will do this as soon as the infrastructure for this is completely working again. This also needs a ticket, will provide the issue number here once it has been created.

### DIFF
--- a/src/all_test.ml
+++ b/src/all_test.ml
@@ -56,6 +56,7 @@ let tlog_tests = "tlogs" >::: [
   Tlogreader2_test.suite;
 ]
 let small_catchup = "small_catchup" >:::[Catchup_test.suite;]
+let store = "store" >:::[Store_test.suite;]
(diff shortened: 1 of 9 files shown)


toolslive:
LGTM

PR #46, opened 2013-05-02
close the tlc after catchup_only

toolslive:
ARAKOON-382:Arakoon fails to start when doing a catchup only before starting


NicolasT:
LGTM

PR #50, opened 2013-05-10
camltc API changed ... needs options

toolslive:


NicolasT:
:-( The required change is in 1.4, forward-merging 1.4 into 1.5 (and 1.5 into 1.6) should have fixed this.

PR #51, opened 2013-05-10
exposing more memory usage statistics

domsj:

PR #52, opened 2013-05-13
Log macro

domsj:
this branch contains 2 changes
- added some memory --statistics
- added a macro for log(_f) statements

I made some whitespace changes to client_protocol too, for that file I suggest adding ?w=1 to the compare url.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,8 +9,11 @@ all: build
 clean:
 	ocamlbuild -clean
 
-build:
+build: build_macros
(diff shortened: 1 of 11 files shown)


NicolasT:
Maybe consider prefixing the new `statistics` with `gc_` or something along those lines, or put them in a new subsection (IIRC statistics can be nested maps)?

domsj:
Hmm, as discussed on skype, the macro idea breaks the crash_logger, so we can't use this branch.
I will reopen the other pull request at https://github.com/Incubaid/arakoon/pull/51 .

PR #55, opened 2013-05-16
Simplify store

domsj:


toolslive:
it's certainly a step in the right direction (LGTM)

PR #56, opened 2013-05-17
get_db on 2 slaves, kill master, see if it comes back up.

toolslive:
maybe there's a cleaner alternative for exposing the protocol than my sys.path trickery.


domsj:
LGTM

PR #57, opened 2013-05-17
Add '-start' no-op flag and use it from extension

NicolasT:
See: ARAKOON-379
See: http://jira.incubaid.com/browse/ARAKOON-379
See: ARAKOON-388
See: http://jira.incubaid.com/browse/ARAKOON-388


toolslive:
I think this works too.

PR #58, opened 2013-05-21
Arakoon 391

domsj:


NicolasT:
Maybe remove the first commit & alter commit message of last?

extension/test/server/right/system_tests_long_right.py
```diff
@@ -347,6 +347,92 @@ def test_drop_master():
...
+            drop_master(n)
+        except Exception, ex:
+            last_ex = ex
+            raise ex
+        busy = False
+
+    t_drop = Thread(target = lambda : inner_drop_master())
+    t_drop.start()
+
+    time.sleep(400)
```
toolslive, line 387: you can use a condition and then wait & notify, but this would work too.

extension/test/server/right/system_tests_long_right.py
```diff
@@ -347,6 +347,92 @@ def test_drop_master():
 """    drop_masters = lambda : drop_master(n)
     Common.create_and_wait_for_threads ( 1, 1, drop_masters, n*8*3 )"""
 
+from threading import Thread
```
toolslive: normally, you put imports at the beginning of the module.

domsj:
Ok, did some history rewriting

PR #61, opened 2013-05-27
Batched local store

domsj:


NicolasT:
TODO:
- [x] Count key/value-sizes inserted in batch, and commit batch when some configurable threshold is reached.
- [x] Make entry count for batch configurable.

src/node/node_main.ml
```diff
@@ -199,7 +199,8 @@ let only_catchup (type s) (module S : Store.STORE with type t = s) ~name ~cluste
   let future_i = Sn.start in
   Catchup.catchup me other_configs ~cluster_id 
     ((module S),store,tlc)  current_i mr_name (future_n,future_i) >>= fun _ ->
-  tlc # close () 
+  tlc # close ();
```
toolslive: tlc # close  () >>= ...

src/node/node_main.ml
```diff
@@ -651,7 +652,7 @@ let _main_2 (type s)
       
       
 let main_t make_config name daemonize catchup_only : int Lwt.t =
-  let module S = (val (Store.make_store_module (module Local_store))) in
+  let module S = (val (Store.make_store_module (module Batched_store.Local_store))) in
```
toolslive, line 667: maybe make this configurable....

src/node/batched_store.ml
```diff
@@ -0,0 +1,279 @@
...
+    S.defrag s.s
+
+  let copy_store s =
+    _sync_and_start_transaction_if_needed s;
+    S.copy_store s.s
+
+  let copy_store2 =
+    S.copy_store2
+
+  let get_fringe s =
```
toolslive, line 272: get_fringe, copy_store, defrag, optimize, get_key_count, get_location,....
all look very similar. maybe it's worth to extract the commonalities.

src/node/batched_store.ml
```diff
@@ -0,0 +1,279 @@
...
+      m
+
+  let _track_or_apply_vo s k vo =
+    match s._ls_tx with
+      | None ->
+          s._current_tx_cache <- StringMap.add k vo s._current_tx_cache
+      | Some ls_tx ->
+          _apply_vo_to_local_store s.s ls_tx k vo
+
+  let _finalize f g =
```
toolslive: unused. (and pretty dangerous )

NicolasT: :-1: This doesn't look OK, better get rid of it.

src/node/batched_store.ml
```diff
@@ -0,0 +1,279 @@
...
+          Lwt.return ()))
+
+  let _verify_tx s tx =
+    match s._tx with
+      | None -> failwith "not in a batched store transaction"
+      | Some tx' ->
+          if tx != tx'
+          then failwith "the provided transaction is not the current transaction of the batched store"
+
+  let exists s k =
```
NicolasT: 'exists' and 'get' share quite some logic. Consider lifting this?

src/node/node_main.ml
NicolasT, line 667: If the number of 'batched items' is configurable, setting this to 1 would yield the old behavior of using plain Local_store, without the need to be able to disable Batched_store. Would this be suitable?

src/node/node_main.ml
NicolasT: Might want to enable warnings-as-errors during build?

src/node/store.ml
```diff
@@ -547,7 +537,12 @@ struct
   let _with_transaction : t -> key_or_transaction -> (transaction -> 'a Lwt.t) -> 'a Lwt.t =
     fun store kt f -> match kt with
       | Key key ->
-          with_transaction store ~key:(Some key) f
+          let matched_locks = match store._tx_lock with
+            | Some txl -> txl == key
```
NicolasT: '==' vs '='?

src/node/store_test.ml
```diff
@@ -125,23 +135,11 @@ let test_safe_insert_value_with_partial_value_update () =
     Lwt.return ()
   )
 
-let test_safe_insert_value_with_tx () =
```
NicolasT, line 128: Care to elaborate why this test was removed?

src/node/batched_store.ml
domsj: Agreed, removing

src/node/batched_store.ml
domsj, line 272: I have extracted (most of) the commonalities already into _sync_and_start_transaction_if_needed ;-).
I don't think any more commonalities are worth the trouble so I'll leave this as is.

src/node/node_main.ml
domsj: after discussion in the chat you're on top of this in a soon to be other pull request :)

src/node/store.ml
domsj: == tests for physical equality, which is what I want here. I'll add a comment to the code to make this more explicit.

src/node/store_test.ml
domsj, line 128: This test verified the behaviour of S.safe_insert_value with a ~tx:(Some a_tx) passed to it. As you can see at https://github.com/Incubaid/arakoon/pull/61/files#L11R818 this possibility has been removed (it was there to allow batching from in catchup.ml).

NicolasT:
Is this ready for merge?

PR #62, opened 2013-05-27
Log defrag administrative commands to info level

NicolasT:
See: ARAKOON-402
See: http://jira.incubaid.com/browse/ARAKOON-402


domsj:
how about also logging (un)quiesce, optimize, get_db to info level?

NicolasT:
Will rework on 1.4, and add logging for other commands.

PR #63, opened 2013-05-27
Bump logging of administrative commands to info level

NicolasT:
See: ARAKOON-402
See: http://jira.incubaid.com/browse/ARAKOON-402

Closes #62.


domsj:
LGTM

PR #65, opened 2013-05-27
Merge 1.4 into 1.5

NicolasT:


NicolasT:
This needs some close review.

domsj:
LGTM, but considering you want close review it might be best if romain has a look too.
Just a small heads up: when forward merging this into 1.6 you'll probably run into the changes to store/local_store, I think (at least part of) the logging moved from local_store to store.

domsj:
I've pushed another commit to the 1.4 branch, do we want to include it into this pull request?

NicolasT:
Will do.

NicolasT:
Merged in latest changes from 1.4.

toolslive:
LGTM, but I would like to see the changes from
- _Lwt.return_ to _FSM.return_
- _some_side_effect >>= rest_ to _Fsm.return ~sides:[some_effect]_
  when things propagate into 1.6

PR #66, opened 2013-05-28
Optimize logging

domsj:
This pull request fixes the crash_logging and implements the optimization in which strings for the crashlogger are only computed when needed.
The lazyness is implemented by the type Crash_logger.msg which is either an immediate string or a (unit -> string).
Some syntactic sugar for this has been added in the form macros.

Logger.{debug/info/...}_f section "format%s string" "bla"
->
Logger.log_ Logger.{Debug/Info/...} section (fun () -> Printf.sprintf "format%s string" "bla")

macros ending with _ also capture the variable 'section' which is expected to be defined in the environment, e.g.:

Logger.{debug/info/...}_f_ "format%s string" "bla"
->
Logger.log_ Logger.{Debug/Info/...} section (fun () -> Printf.sprintf "format%s string" "bla")

I tried to eliminate all usages of Lwt_log (except in the modules specifically dealing with logging ofcourse). Some usages are still left but these are not statements doing logging themselves; if desired I can centralize this into the logger module too.

### DIFF
--- a/logger_macro.ml
+++ b/logger_macro.ml
@@ -1,0 +1,198 @@
+open Camlp4.PreCast
+
+(*
+
+This file defines several macros related to the Logger module.
+
+It is important to keep in mind that macros can't always be used like functions, therefor it's useful to know how these macros are expanded.
+
+The general structure is the following (in which level can be any of Debug|Info|Notice|Warning|Error|Fatal):
+Logger.level    ?exn section msg
+Logger.level_   ?exn msg
+Logger.level_f  ?exn section format_string and args
+Logger.level_f_ ?exn format_string and args
+
+The macros ending with _ grab a variable named section from the enviroment.
+Macros containing _f wrap the 'format_string and args' part into a closure which can at a later time be evaluated to calculate a string.
+
+Some concrete examples:
+
+Logger.debug section "log this"
+=> Logger.log section Logger.Debug "log this"
+
+Logger.debug_ "log this"
+=> Logger.log section Logger.Debug "log this"
+
+Logger.debug_f section "format%s string" "bla"
+=> Logger.log_ section Logger.Debug (fun () -> Printf.sprintf "format%s string" "bla")
+
+Logger.debug_f_ "format%s string" "bla"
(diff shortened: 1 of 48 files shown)


Makefile
```diff
@@ -9,6 +9,7 @@ all: build
 clean:
 	ocamlbuild -clean
 
+
```
NicolasT: Revert?

logger_macro.ml
```diff
@@ -0,0 +1,168 @@
+open Camlp4.PreCast
```
NicolasT, line 1: Might be useful to give an overview of what gets rewritten into what. It's not entirely clear/obvious if you ask me (yet I don't know camlp4, so that might be part of the issue).

src/client/client_protocol.ml
```diff
@@ -62,9 +61,11 @@ let read_command (ic,oc) =
       end
   end
 
+let log_debug m =
+  Logger.debug_ m
```
NicolasT, line 65: Does this still rework as expected when macro rewriting is in place?

Makefile
domsj: I changed this file in 2 different commits, resulting in the net addition of 1 new line.
Do you prefer another commit removing this extra line or should I do some history rewriting and collapse everything into 1 commit (and undo this 1 line in the process)?

logger_macro.ml
domsj, line 1: Ok, I'll add the examples from the explanation in this pull request and expand on them some more.

src/client/client_protocol.ml
domsj, line 65: Yes, this should work as desired.
Macros with a format string can't be expressed like that. It does require the m argument here for it to work.
As mentioned in my comment above, I'll add some more comments to the macro file which should make clearer what will work and what won't.

Makefile
NicolasT: Rewrite some history ;-)

NicolasT:
Next to minor comments, LGTM.

PR #68, opened 2013-05-29
change log level of exceptions with Arakoon.E_NOT_FOUND and Arakoon.E_AS...

domsj:
...SERTION_FAILED from Error to Debug

### DIFF
--- a/src/client/client_protocol.ml
+++ b/src/client/client_protocol.ml
@@ -77,16 +77,16 @@ let response_rc_bool oc rc b =
   Lwt.return false
 
 let handle_exception oc exn=
-  let rc, msg, is_fatal, close_socket = match exn with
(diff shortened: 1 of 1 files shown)


domsj:
FYI: The jira issue only talked about changing the log level of failed asserts, I was opinionated and also changed the log level of Not_found, as to me that's a case of failure of assertSomeValue.

cfg/arakoon.ini
```diff
@@ -1,6 +1,6 @@
 [global]
 #comma separated list of node names that are part of the cluster
-cluster = arakoon_0 
+cluster = arakoon_0 arakoon_1
```
NicolasT: Please undo this.

NicolasT:
LGTM except for the arakoon.ini change.

PR #69, opened 2013-05-29
fix for drop_master_with_load(_and_verify) test

domsj:
This fixes the failing tests related to ARAKOON-391, apparently the fault wasn't in arakoon drop_master implementation but in the way I wrote the test.


NicolasT:
LGTM

PR #70, opened 2013-05-31
optimize logging in multi_paxos.ml, multi_paxos_fsm.ml, multi_paxos_test...

domsj:
....ml and slave.ml

### DIFF
--- a/src/paxos/multi_paxos_fsm.ml
+++ b/src/paxos/multi_paxos_fsm.ml
@@ -39,7 +39,7 @@ let forced_master_suggest constants (n,i) () =
   let n' = update_n constants n in
   mcast constants (Prepare (n',i)) >>= fun () ->
   start_election_timeout constants n >>= fun () ->
-  log_f me "forced_master_suggest: suggesting n=%s" (Sn.string_of n') >>= fun () ->
+  Logger.debug_f_ "%s: forced_master_suggest: suggesting n=%s" me (Sn.string_of n') >>= fun () ->
   let tlog_coll = constants.tlog_coll in
   let l_val = tlog_coll # get_last_value i in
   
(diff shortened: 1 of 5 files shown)


src/paxos/multi_paxos_fsm.ml
```diff
@@ -864,30 +864,28 @@ let rec paxos_produce buffers
...
 
 
-let _execute_effects constants e = 
+let _execute_effects constants e =
   match e with
-    | ELog build ->
-          let s = build () in
-          let s' = "PURE: " ^ s in
-          log constants.me s'
+    | ELog build          -> Logger.debug_f_ "%s: %s" constants.me ("PURE: " ^ build ())
```
NicolasT, line 888: Won't this result in `constants.me` being logged twice, since it's also added to the log string by various `Logger.*` calls?

domsj, line 888: constants.me was previously added in https://github.com/Incubaid/arakoon/pull/70/files#L1L36 , now it's made explicit in the Logger.debug_f_ "%s: previous_format_string" (which is the biggest part of this changeset).
I have however not changed the construction of any ELog effects, so these don't have 'me' in the string being built yet, as it was before. So I don't think constants.me is being logged twice.
(Most ELog statements btw are written with a (fun () -> Printf.sprintf ...) in them, this could be made nicer by introducing a specific macro for this, which I have not done in this change.)

Hmm, now that I think of it, I could've phrased this line as:
-    | ELog build          -> Logger.debug_f_ "%s: PURE: %s" constants.me (build ())
  which would've been nicer but that probably doesn't matter too much :-)

NicolasT, line 888: Ah, right, all other `Logger.*` calls aren't part of `Pure`. LGTM then.

NicolasT:
LGTM :+1:

PR #72, opened 2013-06-03
Investigate invalid operation

domsj:
I recommend viewing this pull request with ?w=1 attached to the url.

I think the exception in http://cislave11.incubaid.com/jenkins/jobs/jenkins-arakoon-1.6-system-tests-slow-RIGHT-83/home/jenkinslxc/qbase3/var/tmp/arakoon_system_tests/test_sso_deployment/sturdy_0/log/sturdy_0.debug.1370002695.000000 came through on_consensus (https://github.com/Incubaid/arakoon/blob/1.6/src/node/store.ml#L850 ) untill https://github.com/Incubaid/arakoon/blob/1.6/src/node/store.ml#L807 where it failed on the _get_j call.
That path doesn't check store.closed (only store.quiesced), so maybe we want to check that there too?
Or is it good enough to die with a message saying the store was closed (added exception wrapping for that)?

### DIFF
--- a/src/node/node_main.ml
+++ b/src/node/node_main.ml
@@ -212,8 +212,9 @@ let only_catchup (type s) (module S : Store.STORE with type t = s) ~name ~cluste
   let future_i = Sn.start in
   Catchup.catchup me other_configs ~cluster_id 
     ((module S),store,tlc)  current_i mr_name (future_n,future_i) >>= fun _ ->
-  tlc # close () >>= fun () ->
-  S.close store
+  S.close store >>= fun () ->
+  tlc # close ()
+
   
     
     
--- a/src/node/node_main.ml
+++ b/src/node/node_main.ml
@@ -603,39 +603,36 @@ let _main_2 (type s)
 	    (fun () ->
           let _ = Lwt_unix.on_signal 15 unlock_killswitch in (* TERM aka kill   *)
           let _ = Lwt_unix.on_signal 2  unlock_killswitch in (*  INT aka Ctrl-C *)
-	      Lwt.pick [
-	        messaging # run ();
-	        begin
-	          build_startup_state () >>= fun (start_state,
-					                          service,
-					                          rapporting) ->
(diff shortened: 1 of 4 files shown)


NicolasT:
I don't get why the store should be closed before the tlog. I'd expect things to be reverse (first close tlog, that's the most important thing, if that fails the store could be broken anyway).

domsj:
Don't we want to know the store has succesfully closed before writing the marker to the tlog (which happens when closing the tlog)?

NicolasT:
Right. I guess some review by @toolslive would be appropriate too.

toolslive:
writing the marker in the tlog is the very last thing we do before the process stops. If we have that, we can be sure everything was ok.

domsj:
another thing occurred to me, maybe we should try to stop the fsm thread first?
because the original reason for the exception actually is that the state machine reaches consensus and tries to manipulate the store...
anyway haven't looked yet how that should be done.

domsj:
To emphasize my previous comment, the following test failed because the fsm was still trying to write to the tlog while it was in the process of closing. As a result of this no marker got written, the wouldn't start anymore and then offcourse it doesn't catch up in a timely fashion.
http://cimaster.incubaid.com/view/arakoon-git/job/arakoon-git-system-tests-slow-LEFT/lastCompletedBuild/testReport/arakoon_system_tests.server.left/system_tests_long_left/test_catchup_exercises/
http://cislave11.incubaid.com/jenkins/jobs/jenkins-arakoon-git-system-tests-slow-LEFT-132/home/jenkinslxc/qbase3/var/tmp/arakoon_system_tests/test_catchup_exercises/sturdy_0/log/sturdy_0.debug.1370435589.000000

I will look into how I can first stop the fsm and then close store and tlog.

domsj:
Ok, CI is good. I believe this is a change for the better.
Please review.

domsj:
I removed the memory leak in server.ml (rewrote some history). The changes succesfully passed CI.

PR #75, opened 2013-06-05
Error-out if BDBFFATAL flag is set on TokyoCabinet database

NicolasT:
When a TokyoCabinet database is opened, the flags contained in the file
are retrieved, and checked for `BDBFFATAL`. If this flag is set, and
exception is thrown which results in the Arakoon process to quit with
return code 45.

For testing purposes, a TokyoCabinet file can be 'broken' manually by
setting bit 2 of the 34th byte in the database file.

See: ARAKOON-393
See: http://jira.incubaid.com/browse/ARAKOON-393

### DIFF
--- a/src/node/local_store.ml
+++ b/src/node/local_store.ml
@@ -28,6 +28,7 @@ open Log_extra
 open Store
 open Unix.LargeFile
 
+exception BdbFFatal of string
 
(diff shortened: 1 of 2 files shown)


domsj:
LGTM

PR #76, opened 2013-06-06
configure batched store from pylabs extensions

domsj:


NicolasT:
Looks good, but I didn't run/test it ;-)

PR #78, opened 2013-06-06
Extension: Pass '-n' to 'pgrep'

NicolasT:
In order not to match any intermediate 'wrapper' scripts, a process
table lookup should only match the 'last' process.

See: ARAKOON-406
See: http://jira.incubaid.com/browse/ARAKOON-406


domsj:
LGTM

PR #80, opened 2013-06-10
introduce difference between cluster's id and cluster's name

toolslive:
basically this is the proposed change. ( 'quick' suite holds)

### DIFF
--- a/extension/client/ArakoonClient.py
+++ b/extension/client/ArakoonClient.py
@@ -178,25 +178,25 @@ class ArakoonClient:
     Arakoon client management
     """
     @staticmethod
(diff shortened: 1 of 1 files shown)


NicolasT:
Can we be certain this won't break any existing code?

toolslive:
it's python... so a definite 'no'.

NicolasT:
It's not really a language thing, rather how things behave ;-)

Anyway, I guess it makes no sense to have an existing setup where the cluster_id doesn't match the cluster_name, or when things work using cluster_name as cluster_id, the cluster_name is also set as cluster_id in the config file, so things should keep working as-is.

LGTM then.

PR #82, opened 2013-06-11
only catch Not_found, Key_not_found and Arakoon_exc.Exception(Arakoon_ex...

domsj:
...c.E_ASSERTION_FAILED, _) exceptions in store

This the change I would want to merge in if we don't want this fixed on the 1.4 branch.


NicolasT:
Looks OK. If CI agrees, LGTM.

PR #85, opened 2013-06-13
only retry catchup connection on another ip if connection never successf...

domsj:
...ully opened

providing last_entries in sync_backend shouldn't fail on TLogCheckSumError

This differs slightly from https://github.com/Incubaid/arakoon/pull/83

### DIFF
--- a/src/node/sync_backend.ml
+++ b/src/node/sync_backend.ml
@@ -414,7 +414,12 @@ object(self: #backend)
           and u = Entry.u_of entry
           in
           Tlogcommon.write_entry oc i u in
-	    tlog_collection # iterate start_i3 too_far_i f >>= fun () ->
+	    Lwt.catch
+          (fun () -> tlog_collection # iterate start_i3 too_far_i f)
+          (function
+            | Tlogcommon.TLogCheckSumError _ -> Lwt.return ()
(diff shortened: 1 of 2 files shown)


src/node/sync_backend.ml
```diff
@@ -414,7 +414,12 @@ object(self: #backend)
           and u = Entry.u_of entry
           in
           Tlogcommon.write_entry oc i u in
-	    tlog_collection # iterate start_i3 too_far_i f >>= fun () ->
+	    Lwt.catch
+          (fun () -> tlog_collection # iterate start_i3 too_far_i f)
+          (function
+            | Tlogcommon.TLogCheckSumError _ -> Lwt.return ()
```
NicolasT, line 420: Might be worth to add a comment why this exception seems to be ignored.

domsj, line 420: Adding a comment would certainly not be a bad idea.
It is indeed ignoring the exception. That exception can only come from inside the tlog#iterate method and not from 'f', so when it happens it's ok to pretend this was the last entry we could supply during catchup.
At this place I don't really care wether the exception indicates a real TLogCheckSumError that aborted the iterate or wether it happens because the last entry in the tlog is concurrently being written and read, I just don't want this to accidently close the connection.

PR #87, opened 2013-06-13
use mutable Hashtbl.t in batched_store

domsj:
Still waiting for CI

### DIFF
--- a/src/node/batched_store.ml
+++ b/src/node/batched_store.ml
@@ -184,15 +184,19 @@ struct
           if tx != tx'
           then failwith "the provided transaction is not the current transaction of the batched store"
 
+  let _find c k =
+    try Some (Hashtbl.find c k)
+    with Not_found -> None
+
   let _with_key_in_caches s k match' else' =
-    if Hashtbl.mem s._current_tx_cache k
-    then
-      match' (Hashtbl.find s._current_tx_cache k)
-    else if Hashtbl.mem s._cache k
-    then
-      match' (Hashtbl.find s._cache k)
-    else
-      else' ()
+    match _find s._current_tx_cache k with
+      | Some v -> match' v
+      | None ->
+          begin
+            match _find s._cache k with
+              | Some v -> match' v
+              | None -> else' ()
+          end
 
   let exists s k =
     _with_key_in_caches s k
--- a/src/node/batched_store.ml
+++ b/src/node/batched_store.ml
@@ -40,8 +40,8 @@ module Batched_store = functor (S : Extended_simple_store) ->
 struct
   type t = {
     s : S.t;
-    mutable _cache : string option StringMap.t;
-    mutable _current_tx_cache : string option StringMap.t;
+    _cache : (string, string option) Hashtbl.t;
+    _current_tx_cache : (string, string option) Hashtbl.t;
 
     _tx_lock : Lwt_mutex.t;
     mutable _tx : transaction option;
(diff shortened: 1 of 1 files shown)


domsj:
CI is good

src/node/batched_store.ml
```diff
@@ -77,19 +77,19 @@ struct
...
     match s._ls_tx with
       | None ->
-          s._current_tx_cache <- StringMap.add k vo s._current_tx_cache
+          Hashtbl.replace s._current_tx_cache k vo
       | Some ls_tx ->
           _apply_vo_to_local_store s.s ls_tx k vo
 
   let _sync_cache_to_local_store s =
-    if not (StringMap.is_empty s._cache)
+    if Hashtbl.length s._cache <> 0
```
NicolasT, line 92: Is there no function to check empty'ness of Hashtbl? Calculating its size just to make sure it's not 0 could be rather heavy.

src/node/batched_store.ml
```diff
@@ -186,12 +185,12 @@ struct
           then failwith "the provided transaction is not the current transaction of the batched store"
 
   let _with_key_in_caches s k match' else' =
-    if StringMap.mem k s._current_tx_cache
+    if Hashtbl.mem s._current_tx_cache k
```
NicolasT: If this change is made for performance reasons, wouldn't it make more sense to
- Try to find the value in the table using 'find'
- Handle exception if not found

instead of going through the map twice (once to check whether the value is in there, once to retrieve it).

src/node/batched_store.ml
```diff
@@ -186,12 +185,12 @@ struct
...
     then
-      match' (StringMap.find k s._current_tx_cache)
-    else if StringMap.mem k s._cache
+      match' (Hashtbl.find s._current_tx_cache k)
+    else if Hashtbl.mem s._cache k
```
NicolasT: Same as above.

src/node/batched_store.ml
domsj, line 92: As far as I can tell from http://caml.inria.fr/pub/docs/manual-ocaml/libref/Hashtbl.html there is no such function.
The Hashtbl.length call however takes constant time according to the documentation, so shouldn't be too bad.

src/node/batched_store.ml
domsj: Guess I'm still used to thinking exceptions are heavy :-).
Ideally I would want find to return a 'b option.
Ok, I'll implement your proposed change.

NicolasT: FWIW, given something which returns an option you can use the "Alternative" pattern (for `'a option`):

``` ocaml
let (<|>) a b = match a with
  | Some _ -> a
  | None -> b ()

let find' h k () =
  try
    let v = Hashtbl.find h k in
    Some v
  with Not_found ->
    None

let foo h1 h2 k f nf =
  let v = find' h1 k () <|> find' h2 k in
  match v with
    | Some v' -> f v'
    | None -> nf ()
```

Or similarly

``` ocaml
let demo h k1 k2 v =
  find' h k1 ()
    <|> find' h k2
    <|> fun () -> Some v
```

src/node/batched_store.ml
toolslive, line 92: from Hashtbl.ml
let length h = h.size

NicolasT:
LGTM?

domsj:
Yes, LGTM too ;-).
Previous version still had some issues, I executed the match' inside the try _ with Not_found, causing me accidentally catch the Not_found exception thrown in match'.
So fixed that in this commit. CI tests are currently running, if these are good (which is my expectation) then all should be good.

PR #88, opened 2013-06-13
remove unnecessary check from tlc2

domsj:
Still waiting for CI

### DIFF
--- a/src/tlog/tlc2.ml
+++ b/src/tlog/tlc2.ml
@@ -472,11 +472,7 @@ object(self: # tlog_collection)
             >>= fun () ->
             self # _rotate file new_outer
           in
-          begin
-            match _previous_entry with
-              | Some pe when (Entry.i_of pe) = i -> Lwt.return file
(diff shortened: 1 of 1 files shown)


domsj:
CI is good

NicolasT:
Why is this 'unnecessary'?

domsj:
I don't see how the check is not yet covered by https://github.com/Incubaid/arakoon/blob/1.6/src/tlog/tlc2.ml#L467 , and neither did romain yesterday.
As far as I understand it either the previous entry is already in the next file, in which case the check on L467 would've been true the previous time but not any more this time (as _outer would be incremented already) or the previous entry is in the previous file, in which case if (Entry.i_of pe) = i then the condition on L467 can't be met.

@toolslive also told me that all edge cases related to tlog handling are under test, so I could remove it and verify with the tests wether my initial assumption (this being unneccessary) was correct.
My hope was that removing it would eventually lead to less confusion when people look at that code again.

toolslive:
LGTM

PR #90, opened 2013-06-17
track ignored threads and cancel them when Lwt_extra.run finishes

domsj:
CI still running

### DIFF
--- a/src/tools/lwt_extra.ml
+++ b/src/tools/lwt_extra.ml
@@ -25,12 +25,6 @@ open Lwt
 
 let igns = Hashtbl.create 10;;
 
-let () =
-  Random.self_init ();
-  Lwt.async_exception_hook :=
-    (function
-      | Lwt.Canceled -> ()
-      | exn -> raise exn)
 
 let rec get_key () =
   let candidate = Random.int 1073741823 in
@@ -42,28 +36,35 @@ let rec get_key () =
 
 let ignore_result (t:'a Lwt.t) =
   let key = get_key () in
-  let t' =
+  let t' () =
     Lwt.finalize
       (fun () -> t)
       (fun () ->
         Hashtbl.remove igns key;
         Lwt.return ()) in
   Hashtbl.add igns key t;
-  Lwt.ignore_result t';;
-
+  Lwt.ignore_result (t' ())
 
 
 let run t =
-  let act =
+  let act () =
     Lwt.finalize
       (fun () -> t)
       (fun () ->
         Lwt.catch
           (fun () ->
+            let async_exc_hook = !Lwt.async_exception_hook in
+            Lwt.async_exception_hook :=
+              (function
+                | Lwt.Canceled -> ()
+                | exn -> async_exc_hook exn);
             let ignored_threads = Hashtbl.fold (fun k t acc -> t :: acc) igns [] in
             Hashtbl.reset igns;
-            Lwt.pick (Lwt.return () :: ignored_threads) >>= fun a ->
-            Lwt.return ())
-          (fun exn -> Lwt.return ()))
+            Lwt.finalize
+              (fun () -> Lwt.pick (Lwt.return () :: ignored_threads))
+              (fun () -> Lwt.async_exception_hook := async_exc_hook; Lwt.return ()))
+          (fun exn ->
+            Logger.info Logger.Section.main ~exn "Exception while cleaning up ignored threads"))
   in
-  Lwt_main.run act
+  Lwt_main.run (act ())
+
--- a/src/tools/lwt_extra.ml
+++ b/src/tools/lwt_extra.ml
@@ -23,24 +23,47 @@ If not, see <http://www.gnu.org/licenses/>.
 open Lwt
 
 
-let collapse_remote ip port cluster_id n = 
-  let t () = 
-    begin
-      if n < 1 
-      then Lwt.fail (Failure ("n should be >= 1"))
-      else Lwt.return ()
-    end >>= fun () ->
-    let address = Network.make_address ip port in
-    let collapse conn =
-      Remote_nodestream.make_remote_nodestream cluster_id conn
-      >>= fun (client:Remote_nodestream.nodestream) ->
-      client # collapse n >>= fun () ->
-      Lwt.return 0 
-    in
-    Lwt.catch
-      (fun () -> Lwt_io.with_connection address collapse)
-      (fun exn -> Logger.fatal Logger.Section.main ~exn "remote_collapsing_failed" 
-	>>= fun () -> Lwt.return (-1)
-      )
+let igns = Hashtbl.create 10;;
+
+let () =
+  Random.self_init ();
+  Lwt.async_exception_hook :=
+    (function
+      | Lwt.Canceled -> ()
+      | exn -> raise exn)
+
+let rec get_key () =
+  let candidate = Random.int 1073741823 in
+  if Hashtbl.mem igns candidate
+  then
+    get_key ()
+  else
+    candidate
+
+let ignore_result (t:'a Lwt.t) =
+  let key = get_key () in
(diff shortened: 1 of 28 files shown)


NicolasT:
See Skype-feedback about `run` implementation.

NicolasT:
If CI and @toolslive agree, LGTM.

domsj:
So far CI ain't agreeing, still making some changes :)

src/tools/lwt_extra.ml
```diff
@@ -0,0 +1,69 @@
...
+GNU Affero General Public License along with this program (file "COPYING").
+If not, see <http://www.gnu.org/licenses/>.
+*)
+
+open Lwt
+
+
+let igns = Hashtbl.create 10;;
+
+let () =
```
NicolasT: Not sure having side-effects to be executed 'at module loading time' is a good idea (and seems unnecessary here, read on)

src/tools/lwt_extra.ml
```diff
@@ -0,0 +1,69 @@
...
+  Random.self_init ();
```
NicolasT: No need to init the PRNG. All we need is some pseudo-random numbers, no need to have some impure seed or whatever.

src/tools/lwt_extra.ml
```diff
@@ -0,0 +1,69 @@
...
+  Lwt.async_exception_hook :=
```
NicolasT: I'd wrap setting and restoring `async_except_hook` in the `run` implementation, so normal Lwt behavior is kept.

src/tools/lwt_extra.ml
```diff
@@ -0,0 +1,69 @@
...
+    (function
+      | Lwt.Canceled -> ()
```
NicolasT: Maybe log something instead?

src/tools/lwt_extra.ml
```diff
@@ -0,0 +1,69 @@
...
+let ignore_result (t:'a Lwt.t) =
+  let key = get_key () in
+  let t' =
+    Lwt.finalize
+      (fun () -> t)
+      (fun () ->
+        Hashtbl.remove igns key;
+        Lwt.return ()) in
+  Hashtbl.add igns key t;
+  Lwt.ignore_result t';;
```
NicolasT: No need for ;;

src/tools/lwt_extra.ml
```diff
@@ -0,0 +1,69 @@
...
+
+let ignore_result (t:'a Lwt.t) =
+  let key = get_key () in
+  let t' =
+    Lwt.finalize
+      (fun () -> t)
+      (fun () ->
+        Hashtbl.remove igns key;
+        Lwt.return ()) in
+  Hashtbl.add igns key t;
```
NicolasT, line 45: Consider adding this _inside_ the action contained in `Lwt.finalize` (i.e. on line 47).

Rationale: what if `Lwt_main.run` returns after the key & action are added to the map, but before the next line is reached? They should only be in the map if the action has actually been launched.

src/tools/lwt_extra.ml
```diff
@@ -0,0 +1,69 @@
...
+
+
+let run t =
+  let act =
+    Lwt.finalize
+      (fun () -> t)
+      (fun () ->
+        Lwt.catch
+          (fun () ->
+            let ignored_threads = Hashtbl.fold (fun k t acc -> t :: acc) igns [] in
```
NicolasT, line 79: I'm surprised this doesn't generate an 'unused binding' compiler warning for `k`.

src/tools/lwt_extra.ml
```diff
@@ -0,0 +1,69 @@
...
+            Hashtbl.reset igns;
+            Lwt.pick (Lwt.return () :: ignored_threads) >>= fun a ->
+            Lwt.return ())
+          (fun exn -> Lwt.return ()))
```
NicolasT: Consider logging the exception?

src/tools/lwt_extra.ml
domsj: Ok

src/tools/lwt_extra.ml
domsj: I can't use Logger.bla here as these return unit Lwt.t
Also, these cancelleds are actually to be expected

src/tools/lwt_extra.ml
domsj, line 45: Ok, considered this.

I want to add the k,v pair to the hashtbl before other threads get the chance to run and possibly ignore another thread, for which the random number generator would return the same number (ok, chances for this are very very slim, but then maybe I should also remove the logic in get_key?)

As for the what if, the Lwt_main.run can only finish when it receives control again (e.g. when a Lwt.bind happens). This can only happen after the key has been added and the thread has successfully been ignored (and possibly even line 47 would at that time not yet be executed). So I'll leave this as is.

src/tools/lwt_extra.ml
domsj, line 79: Yea, sorta suprising. Maybe we need an extra compiler flag to trigger warnings for this?

src/tools/lwt_extra.ml
domsj: Ok

src/tools/lwt_extra.ml
NicolasT, line 45: OK

src/tools/lwt_extra.ml
```diff
@@ -0,0 +1,92 @@
...
+  let t' () =
+    Lwt.finalize
+      (fun () -> t)
+      (fun () ->
+        Hashtbl.remove igns key;
+        Lwt.return ()) in
+  Hashtbl.add igns key t;
+  Lwt.ignore_result (t' ())
+
+let detach (t: unit -> string) =
```
NicolasT, line 48: Is the type restriction required?

domsj, line 48: As it is implemented right now, yes, otherwise the hashtbl above (https://github.com/Incubaid/arakoon/pull/90/files#L26R27) has an 'a in its type which the compiler complains about.
So for now only detached computations returning strings are supported. I think each return type needs it's own Hashtbl tracking the openstanding threads? Or maybe some other clever implementation :), will think if I can find a better way.

src/tools/lwt_extra.ml
```diff
@@ -0,0 +1,92 @@
...
+  let t' = Lwt_preemptive.detach t () in
+  let key = get_key () in
+  Hashtbl.add detacheds key t';
+  Lwt.finalize
+    (fun () -> t')
+    (fun () ->
+      Hashtbl.remove detacheds key;
+      Lwt.return ())
+
+let async_exception_hook : (exn -> unit) ref =
```
NicolasT, line 58: I don't understand why a ref is used here.

src/tools/lwt_extra.ml
```diff
@@ -0,0 +1,92 @@
...
+    prerr_string "Fatal error: exception ";
+    prerr_string (Printexc.to_string exn);
+    prerr_char '\n';
+    Printexc.print_backtrace stderr;
+    flush stderr;
+    exit 2)
+let () =
+  Lwt.async_exception_hook :=
+    (function
+      | Lwt.Canceled -> ()
```
NicolasT, line 69: I'm (still) not convinced it's a good idea to have this as the default. `Lwt.Canceled` in threads during normal operation is a bug. Only when `run` finalizes, these things are expected to occur.

src/tools/lwt_extra.ml
domsj, line 58: It's the same as here: https://github.com/avsm/lwt/blob/master/src/core/lwt.ml#L179
The idea is that this can be set/configured from anywhere else statically.
Another option would be to pass in the async_exception_hook as a parameter of run.

PR #97, opened 2013-06-24
separated shaky tests into separate system_tests_long_shaky test suite (...

domsj:
...test_collapse, test_concurrent_collapse_fails, test_catchup_exercises, test_restart_master_long, test_243, test_272, test_inject_as_head)

CI has already been updated (and tested on arakoon-git)


NicolasT:
LGTM if CI is OK.

One thing crossed my mind: maybe instead of putting all shaky tests in a single module/file, use one per test?

PR #99, opened 2013-06-25
don't die brutally when something goes wrong in either the messaging or ...

domsj:
...fsm thread


domsj:
base tests seems to hang on CI after this change.. need to investigate

NicolasT:
These commits seem to do more than what the PR title suggests. Intentional? LGTM otherwise.

domsj:
I noticed test_db_defrag failed on this branch (it is unrelated to my change though), so I decided to move it to the shaky suite too. The way it failed (with the kill -9) some information I needed to really analyze the issue was no longer available, so that's why I made the third commit.
It would indeed have been clearer to do this on a separate branch though...

PR #101, opened 2013-06-25
return information about how the node(s) got killed

toolslive:
and use it in the system tests


domsj:
LGTM
didn't we also want to check tlog marker presence on teardown? otoh tlog marker should be there if the node wasn't brutally killed with a -9 ...

PR #104, opened 2013-06-26
bump log level of some messages while shutting down

domsj:
Forgot to follow this suggestion from Nicolas in an earlier branch/commit


NicolasT:
These changes seem to do quite a bit more than what the PR title suggests. Is this intentional?

PR #107, opened 2013-06-27
Romain

toolslive:
should fix ARAKOON-386

### DIFF
--- a/cfg/arakoon.ini
+++ b/cfg/arakoon.ini
@@ -1,6 +1,6 @@
 [global]
 #comma separated list of node names that are part of the cluster
-cluster = arakoon_0 ,arakoon_1, arakoon_2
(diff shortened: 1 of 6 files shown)


cfg/arakoon.ini
```diff
@@ -1,6 +1,6 @@
 [global]
 #comma separated list of node names that are part of the cluster
-cluster = arakoon_0 ,arakoon_1, arakoon_2
+cluster = arakoon_0 ,arakoon_1
```
domsj, line 3: Is this change intentional?

toolslive, line 3: does it really matter?

domsj:
If CI agrees then it LGTM.

cfg/arakoon.ini
domsj, line 3: no

PR #108, opened 2013-06-27
fix concurrent collapse test

domsj:
While analyzing the log of this failed test I noticed that instead of being concurrent the collapses came after each other. This change should fix that.


toolslive:
k.

PR #111, opened 2013-06-28
fix UnboundLocalError: local variable 'client' referenced before assignm...

domsj:
...ent seen in test_drop_master_with_load_and_verify


NicolasT:
LGTM

PR #112, opened 2013-07-01
log %s is master on consensus of a new master

domsj:


NicolasT:
Was there a reason for the old log message to be removed?

domsj:
I think it's something romain just forgot in https://github.com/Incubaid/arakoon/commit/163ffecf0458276512bc5296609ec36ffe2d4d62 , I don't see a real reason.

domsj:
updated (rewrote history with new commit)

NicolasT:
LGTM

PR #115, opened 2013-07-02
inject client update event in ignored thread so it can't make the paxos fsm hang

domsj:
CI looks good so far (3 runs of the drop master tests)


toolslive:
LGTM.

PR #116, opened 2013-07-03
various leftovers

toolslive:
-j 4 in makefile speeds up build
debian/changelog so we build 1.6 debian packages
mp_msg shows more from values if we log them


domsj:
LGTM

PR #118, opened 2013-07-03
Nose tests

domsj:
Still running these on CI to see if I made any stupid mistakes.
Some of the reenabled tests (will) fail, nonetheless I would like this to be integrated, as that will create the impetus to fix these.

Will post an update when CI looks good (with the caveat above).


domsj:
replaced by #173

PR #122, opened 2013-07-12
throw TLogUnexcpectedEndOfFile when reaching the end of the tlog unexpec...

domsj:
...tedly, and treat as a TLogCheckSumError in most places

This fixes a bug in truncating introduced in 2f2371e305f677a17c4c19e32b82f38fdc5b9369


NicolasT:
I'm not very familiar with this codebase, but didn't see any obviously hairy things, so if you feel confident & CI agrees, OK to merge.

PR #123, opened 2013-07-15
added head_dir configuration option

domsj:
CI is good


src/node/catchup_test.ml
```diff
@@ -113,7 +113,7 @@ let setup () =
     
 let test_common () =
   Logger.info_ "test_common" >>= fun () ->
-  Tlc2.make_tlc2 _dir_name _tlf_dir true false "node_name" >>= fun tlog_coll ->
+  Tlc2.make_tlc2 _dir_name _tlf_dir _tlf_dir true false "node_name" >>= fun tlog_coll ->
```
NicolasT, line 116: No change require, but I'd use something like

``` ocaml
let head_dir = _tlf_dir in
Tlc2.make_tlc2 _dir_name _tlf_dir head_dir ...
```

for clarity.

NicolasT:
Assuming this passes CI (as reported on Skype), LGTM.

PR #125, opened 2013-07-15
always propagate TLogCheckSumError and TLogUnexpectedEndOfFile exception...

domsj:
...s

Still verifying changes with CI, but manual testing of the behaviour looked good.


domsj:
CI looks good

PR #126, opened 2013-07-18
open and verify tlog before opening the store so that arakoon can exit w...

domsj:
...ith the desired exit codes

Waiting for CI


NicolasT:
LGTM if CI is happy (and I'd expect it to be, otherwise something very unexpected is going on).

PR #127, opened 2013-07-18
Dump 'crash log' on USR2

NicolasT:
For debugging and investigation purposes, whenever SIGUSR2 is sent to
and Arakoon process, it will dump the content of its 'crash logger',
similar to what happens when a node dies.


NicolasT:
Validated manually. Should run through CI?

domsj:
LGTM.
I see no point in running it through CI, so go ahead and merge it in.

PR #128, opened 2013-07-18
Pylabs: Dump crash log before using SIGKILL

NicolasT:
For investigation purposes, it's useful to have access to the node
crash log content when needing to kill it using `SIGKILL` (if it
doesn't quit after `SIGTERM` in time).

With this change the extension `stop` method sends `SIGUSR2` to the
process before sending `SIGKILL`.

See: 7f77f2fd6b9733affb0f9ecadbbcb75ec8e242b7
See: #127

### DIFF
--- a/extension/server/ArakoonManagement.py
+++ b/extension/server/ArakoonManagement.py
@@ -32,6 +32,7 @@ import itertools
 import subprocess
 import time
 import types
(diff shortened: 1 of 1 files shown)


domsj:
LGTM

PR #129, opened 2013-07-18
Arakoon 416

domsj:


src/node/batched_store.ml
```diff
@@ -250,9 +250,15 @@ struct
...
-  let close s =
+  let flush s =
     _sync_and_start_transaction_if_needed s;
-    S.close s.s
+    Lwt.return ()
+
+  let close s flush =
+    if flush
+    then
+      _sync_and_start_transaction_if_needed s;
```
NicolasT, line 260: Can use `flush s` here?

src/tools/server.ml
```diff
@@ -122,10 +122,10 @@ let make_server_thread
                   Lwt.return()
           in
           let t = client_thread () in
-          Hashtbl.add client_threads cid t;
           Lwt.ignore_result
             (Lwt.finalize
                (fun () ->
+                 Hashtbl.add client_threads cid (t, fd);
```
NicolasT, line 128: Why is this moved?

src/tools/server.ml
```diff
@@ -165,14 +165,33 @@ let make_server_thread
           setup_callback () >>= fun () ->
           server_loop ())
         (fun () ->
-          let cancel t =
+          Lwt.catch
+            (fun () ->
+              Logger.debug_ "closing listening socket" >>= fun () ->
+              Lwt_unix.close listening_socket >>= fun () ->
+              Logger.debug_ "closed listening socket")
+            (fun exn -> Lwt.return ()) >>= fun () ->
```
NicolasT: Log?

NicolasT:
Some minor nitpicks and questions, and I'd squash  3f054c9 and  767819b, but overall nothing major.

src/node/batched_store.ml
domsj, line 260: Do you mean S.flush s.s?
That wouldn't flush the batched up changes which is what I want here.

src/tools/server.ml
domsj, line 128: I wanted to make sure the Hashtbl only tracks threads when they are in the scope where they'll also be removed once they finish/are canceled (which is in the finalize clause).

src/tools/server.ml
domsj: Ok

PR #130, opened 2013-07-19
Arakoon 416 tlogcompression

domsj:
Waiting for CI

### DIFF
--- a/src/tlog/tlc2.ml
+++ b/src/tlog/tlc2.ml
@@ -33,6 +33,7 @@ exception TLCNotProperlyClosed of string
 let section = Logger.Section.main
 
 let file_regexp = Str.regexp "^[0-9]+\\.tl\\(og\\|f\\|c\\)$"
+let tlog_regexp = Str.regexp "^[0-9]+\\.tlog$"
 let file_name c = Printf.sprintf "%03i.tlog" c
 let archive_extension = ".tlf"
 let archive_name c = Printf.sprintf "%03i.tlf" c
@@ -365,9 +366,8 @@ object(self: # tlog_collection)
   val mutable _outer = new_c (* ~ pos in dir *)
   val mutable _previous_entry = last
   val mutable _compression_q = Lwt_buffer.create_fixed_capacity 5
-  val mutable _compressing = false
+  val mutable _compression_thread = None
   val _write_lock = Lwt_mutex.create ()
-  val _jc = (Lwt_condition.create () : unit Lwt_condition.t)
   initializer self # start_compression_loop ()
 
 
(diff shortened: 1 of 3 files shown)


src/tlog/tlc2.ml
```diff
@@ -394,48 +395,86 @@ object(self: # tlog_collection)
...
     in
-    Lwt.ignore_result (loop ())
+
+    let add_previous_compression_jobs () =
+      File_system.lwt_directory_list tlog_dir >>= fun entries ->
+      let filtered = List.filter
+        (fun e -> Str.string_match tlog_regexp e 0)
+        entries
+      in
+      let my_compare fn1 fn2 =
```
NicolasT: This and the lines below look like a really complicated way to achieve what you want (I think).

What about (OTOH):

``` ocaml
(* Sort oldest-to-newest *)
let sorted = List.sort (fun a b -> compare (get_number a) (get_number b)) filtered in

let rec add_jobs = function
  | [] -> return ()
  | [_] -> return () (* Skip last *)
  | (hd :: tl) -> add_compression_job (get_number hd) >>= fun () -> add_jobs tl
in
add_jobs sorted
```

src/tlog/tlc2.ml
```diff
@@ -394,48 +395,86 @@ object(self: # tlog_collection)
...
+      let sorted = List.sort my_compare filtered in
+      match sorted with
+        | [] -> Lwt.return ()
+        | hd :: tl ->
+            Lwt_list.iter_s
+              (fun old_tlog ->
+                self # _add_compression_job (get_number old_tlog))
+              (List.rev tl)
+    in
+    Lwt.ignore_result (add_previous_compression_jobs ());
```
NicolasT, line 467: Must this be in an ignore_result? I guess this could be done 'inline' at startup instead?

src/tlog/tlc2.ml
```diff
@@ -612,36 +652,38 @@ object(self: # tlog_collection)
...
-      end >>= fun () ->
-      F.close file
-    end
+      match _compression_thread with
+        | None -> assert false
+        | Some t ->
+            Logger.debug_f_ "Cancelling compression thread" >>= fun () ->
+            (try
+              Lwt.cancel t;
+            with exn -> ());
```
NicolasT: Log something? This seems unexpected to me.

src/tlog/tlc2_test.ml
```diff
@@ -230,6 +231,61 @@ let test_compression_bug (dn, tlf_dir, factory) =
...
+    Compression.uncompress_tlog tlfpath (dn ^ "/" ^ tlf ^ ".tlog") >>= fun () ->
+    File_system.unlink tlfpath
+  in
+  let compresseds = ["000"; "001"; "002"; "003"] in
+  Lwt_list.iter_s uncompress compresseds >>= fun () ->
+  File_system.lwt_directory_list dn >>= fun tlog_entries ->
+  assert ((List.length tlog_entries) = 5);
+
+  let touch fn =
+    Lwt_unix.system (Printf.sprintf "touch %s" fn) >>= fun _ ->
```
NicolasT: Seriously?

NicolasT:
Some code-wise remarks, but overall LGTM

src/tlog/tlc2.ml
domsj: Agreed

src/tlog/tlc2.ml
domsj, line 467: Where is inline at startup?
The compression loop is started from the object initializer, I'm not complete sure if returning it a unit Lwt.t is allowed, and even so it wouldn't run the thread.
If you imply to join it together with ignored compression loop, then this would have to be before the loop, and this could possibly result in deadlocking on adding items to the queue (as there is no process yet taking stuff off).

src/tlog/tlc2.ml
domsj: Ok

src/tlog/tlc2.ml
NicolasT, line 467: Ooooh, an initializer with such side-effect, hmh... I always consider that to be a smell. How about adding some 'initialize' method which spawns of the compression loop, then does the file listing & filling the queue (not in a thread, but as-is), and as such returns `unit Lwt.t`?

PR #131, opened 2013-07-19
Bump all logging to debug level after SIGTERM/SIGINT

NicolasT:
For investigation & debugging purposes it's useful to know what happens
after a node receives SIGTERM or SIGINT (e.g. to figure out why it's
taking a long time to quit). This change makes sure all logging is
bumped to debug level, ignoring any prior configuration.

See: ARAKOON-416
See: http://jira.incubaid.com/browse/ARAKOON-416


NicolasT:
Validated manually (by first configuring the node to log at error level only), seems to work as expected.

domsj:
LGTM

PR #133, opened 2013-07-22
don't swallow exceptions so the crashlog can be triggered

domsj:


src/node/node_main.ml
```diff
@@ -578,7 +578,9 @@ let _main_2 (type s)
           let log_exception m t =
             Lwt.catch
               t
-              (fun exn -> Logger.warning_ ~exn m) in
+              (fun exn ->
+                Logger.warning_ ~exn m >>= fun () ->
```
NicolasT: Maybe bump level above warning?

domsj: Ok.
It will also still log the exception at fatal in https://github.com/Incubaid/arakoon/blob/2d7bfb336508c8d084b0de5dbac30e2658124f5b/src/node/node_main.ml#L642 , but that of course loses the info about which thread the exception happened in + it happens later, by which time the node might be hanging :). New commit coming up.

PR #134, opened 2013-07-22
add a compression job for the previous tlog if the first value being log...

domsj:
...ged starts a new one (fixes test_sabotage)


NicolasT:
LGTM if CI is happy.

PR #135, opened 2013-07-22
inject FromClient into correct buffer

domsj:
This should fix ARAKOON-420 and ARAKOON-421.

The nodes die because they receive a FromClient message in the slave_wait_for_accept or wait_for_promises state.
A quick look at https://github.com/Incubaid/arakoon/blob/1.6/src/paxos/multi_paxos_fsm.ml#L739 and https://github.com/Incubaid/arakoon/blob/1.6/src/paxos/multi_paxos_fsm.ml#L732 shows that these states cannot receive those messages as they are not subscribed to the messages those buffers should be in.
The Nop inserted at https://github.com/Incubaid/arakoon/blob/1.6/src/paxos/master.ml#L43 however goes through the inject_event which is fixed in this pull request.

### DIFF
--- a/src/node/node_main.ml
+++ b/src/node/node_main.ml
@@ -500,16 +500,24 @@ let _main_2 (type s)
 	      let get_last_value (i:Sn.t) = tlog_coll # get_last_value i in
 	      let election_timeout_buffer = Lwt_buffer.create_fixed_capacity 1 in
 	      let inject_event (e:Multi_paxos.paxos_event) =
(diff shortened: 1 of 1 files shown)


NicolasT:
Guess I need to trust your judgement on this one :smiley:

domsj:
I hoped it would be crystal clear with my lovely explanation, relevant parts are linked so you can easily verify! :)

PR #137, opened 2013-07-26
Arakoon 359

domsj:
TODO
- fleshing out exact requirements with amplidata
- get_db, collapsing, ... ?
- python extensions code
- some tests

But feel free to have a look already and provide some comments

### DIFF
--- a/cfg/arakoon.ini
+++ b/cfg/arakoon.ini
@@ -74,6 +74,7 @@ home = /tmp/arakoon/arakoon_1
 log_level = info
 log_config = default_log_config
 batched_transaction_config = default_batched_transaction_config
+quiesced = true
 
 
 [arakoon_2]
--- a/src/node/node_main.ml
+++ b/src/node/node_main.ml
@@ -442,13 +442,6 @@ let _main_2 (type s)
           make_tlog_coll me.tlog_dir me.tlf_dir me.head_dir me.use_compression me.fsync name
           >>= fun (tlog_coll:Tlogcollection.tlog_collection) ->
           S.make_store db_name >>= fun (store:S.t) ->
-          begin
-            if me.is_quiesced
-            then
-              S.quiesce store
-            else
-              Lwt.return ()
-          end >>= fun () ->
           let last_i = tlog_coll # get_last_i () in
           let ti_o = Some last_i in
           let current_i = (* confusingly ~current_i further in the callstack is too_far_i *)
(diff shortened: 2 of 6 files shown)


cfg/arakoon.ini
```diff
@@ -74,6 +74,7 @@ home = /tmp/arakoon/arakoon_1
 log_level = info
 log_config = default_log_config
 batched_transaction_config = default_batched_transaction_config
+quiesced = true
```
NicolasT, line 77: I'd keep this commented-out by default.

NicolasT, line 77: Maybe we should find a better name. `quiesced` might match the implementation, but isn't obvious from a user perspective.

src/node/node_main.ml
```diff
@@ -451,6 +451,17 @@ let _main_2 (type s)
               Sn.succ last_i
             else
               last_i in
+          begin
+            if me.is_quiesced
+            then
+              begin
+                S.quiesce store >>= fun () ->
+                S._set_i store (Sn.pred last_i);
```
NicolasT, line 459: Why's this?

src/node/node_main.ml
```diff
@@ -559,19 +570,26 @@ let _main_2 (type s)
...
-		        end
-	        | _ -> Multi_paxos_fsm.enter_simple_paxos
-	    in
+        let to_run =
+          match master with
+          | Forced master  ->
+            if master = my_name
+            then Multi_paxos_fsm.enter_forced_master
+            else
+              begin
```
NicolasT, line 579: If I see it right, this code is OK when a node is quiesced in a forced-master setup, but it's not immediately obvious. Maybe make it an explicit case in here as well?

domsj, line 579: Hmm yes, might be better explicit.
Forced master .. that actually means all other nodes are forced slaves, in which case they might as well all be quiesced too, right?
So maybe the name for quiesced could be changed into forced slave, which as an optimisation just happens to not keep a nodename.db around ? (I think this should be considered when meeting with amplidata about requirements.)

src/node/node_main.ml
domsj, line 459: This fakes that the i of the store is quite advanced already, so that when a quiesced node is restarted, which has already accumulated quite a few .tlf/.tlog files, it doesn't have to replay all those tlfs/tlogs.
I added this in the second commit ("optimize catchup of store during startup phase of a quiesced node ") but I admit it could've used a comment here.

domsj:
replaced by #175

PR #138, opened 2013-07-29
remove erroneous assert in tlog rotate

domsj:
rename rotate to jump_to_tlog to better indicate it's true nature

Still running it on CI, but this shouldn't break anything.


toolslive:
K.

PR #141, opened 2013-08-07
treat promises for MasterSets as if nodes are promising they want me as ...

domsj:
...the new master

fix return value of verify_n_catchup_store

Still testing it on CI


toolslive:
LGTm

PR #142, opened 2013-08-08
Arakoon 424

toolslive:
actually ARAKOON-425.

### DIFF
--- a/extension/test/server/system_tests_common.py
+++ b/extension/test/server/system_tests_common.py
@@ -440,9 +440,17 @@ def regenerateClientConfig( cluster_id ):
 def wipe(name):
     config = getConfig(name)
     data_dir = config['home']
-    q.system.fs.removeDirTree(data_dir)
-    q.system.fs.createDir(data_dir)
-    logging.info("wiped %s" % name)
+    dirs = [data_dir]
+    def wipe_dir(d):
+        q.system.fs.removeDirTree(d)
+        q.system.fs.createDir(d)
+    wipe_dir(data_dir)
+    tlf_dir = config.get('tlf_dir')
+    if tlf_dir:
+        wipe_dir(tlf_dir)
+        dirs.append(data_dir)
+
+    logging.info("wiped %s (dirs=%s)",name, str(dirs))
 
 def get_memory_usage(node_name):
     cluster = _getCluster()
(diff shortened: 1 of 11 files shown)


extension/test/server/system_tests_common.py
```diff
@@ -440,9 +440,17 @@ def regenerateClientConfig( cluster_id ):
...
-    logging.info("wiped %s" % name)
+    dirs = [data_dir]
+    def wipe_dir(d):
+        q.system.fs.removeDirTree(d)
+        q.system.fs.createDir(d)
+    wipe_dir(data_dir)
+    tlf_dir = config.get('tlf_dir')
+    if tlf_dir:
+        wipe_dir(tlf_dir)
+        dirs.append(data_dir)
```
domsj: is this meant to be dirs.append(tlf_dir) ?

src/node/catchup.ml
```diff
@@ -327,3 +329,172 @@ let get_db db_name cluster_id cfgs =
     end
   in
   Lwt_list.fold_left_s try_db_download None cfgs 
+
+
+
+
+let last_entries
```
domsj, line 336: there was an opportunity here to fix the indentation (otoh in a few versions this code will probably be thrown out)

src/node/node_main.ml
```diff
@@ -209,14 +209,20 @@ let only_catchup (type s) (module S : Store.STORE with type t = s) ~name ~cluste
...
-  Catchup.catchup me.Node_cfg.Node_cfg.node_name other_configs ~cluster_id 
-    ((module S),store,tlc)  current_i mr_name (future_n,future_i) >>= fun _ ->
-  S.close store >>= fun () ->
-  tlc # close ()
+  make_tlog_coll me.tlog_dir me.tlf_dir me.head_dir 
+    me.use_compression me.fsync name >>= fun tlc ->
+  Lwt.finalize
+    (fun () ->
+      let current_i = match S.consensus_i store with
+        | None -> Sn.start
```
domsj, line 217: So a store with 1 entry and an empty store get the same current_i; that can't be right.
I know it happens like this in other places in the code too, so no need to change it here, but we should pick it up sometime...

domsj:
otherwise LGTM

PR #144, opened 2013-08-09
when discovering a new master take their n

domsj:


NicolasT:
Seems sensible.

domsj:
Merged this into 1.7

PR #147, opened 2013-08-12
Shorter timeouts

domsj:
CI is ok


domsj:
The idea is that these shorter timeouts should expose more bad behaviour (if any)

toolslive:
have fun ;)

PR #148, opened 2013-08-12
insert ElectionTimeout before going to Accepteds_check_done

domsj:
refactoring in Master_consensus to reuse an existing code path

CI looks good, but tests took longer to run than expected.
So I've launched them again to see if they finish quicker now... will post an update when these finish.

Might be caused by having all these extra timeout threads now? (But we need them .. or we need another way to schedule these events, maybe also allow cancelling of previous timeout etc...)
Any comments/suggestions more than welcome :)


domsj:
Changed how ElectionTimeouts are handled and timings on CI are normal now.

toolslive:
LGTM

PR #149, opened 2013-08-12
forced masters should not respond to clients when they're not the master...

domsj:
...,

as this could mean some updates are not yet applied to the store

This fixes a bug where when restarting the forced master it does not yet apply the latest value from it's tlog to the store before it starts responding to client requests. This too eager responding might return outdated data (and resulted in a failure on CI).

Ran it on CI, looks good, except for 1 bug, which I will not solve in this PR.
I have a fix for it already on the branch timeout-accepteds-check-done which should hopefully soon get a PR too (if it doesn't I'll apply the fix on a new separate branch).

### DIFF
--- a/src/node/sync_backend.ml
+++ b/src/node/sync_backend.ml
@@ -382,7 +382,7 @@ struct
             | None -> None,"young cluster"
             | Some (m,ls) ->
               match Node_cfg.get_master cfg with
-                | Elected | Preferred _ ->
+                | Elected | Preferred _ | Forced _ ->
                   begin
                     if (m = my_name ) && (ls < instantiation_time)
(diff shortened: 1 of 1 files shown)


domsj:
matching _ is in general a bad idea

domsj:
When adding a new constructor to the variant the compiler will point out where you're not matching that constructor, unless of course you match _. Matching _ means you give up some of the benefits of static typing.
Therefor 'in general' I think it's better to be more explicit so that when you add new constructors the compiler can force you think about it. There will probably be cases where you can argue the variant won't extended, or where the new constructors behaviour should always match the _ catchall in which case using _ is thus a good thing to do, but I tend to err on the safe side.

PR #151, opened 2013-08-12
respect another node running for master by temporarily not sending promi...

domsj:
...ses to other nodes

### DIFF
--- a/src/paxos/multi_paxos.ml
+++ b/src/paxos/multi_paxos.ml
@@ -288,16 +288,33 @@ let handle_prepare (type s) constants dest n n' i' =
           else
             begin
               (* Ok, we can make a Promise to the other node, if we want to *)
-              let show_respect = match constants.respect_run_master with
-                | None -> false
+              let make_promise () =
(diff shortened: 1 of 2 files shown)


NicolasT:
Clever, although I think this might cause issues in e.g. a 5-node setup (N1/2/3/4/5) where N1 can talk to N2 but not any other, N2/3/4 can all talk, and nobody can talk to N5 (and the other way around). If N1 is 'fast enough' to send messages to N2, won't things get stuck (for at least a while)?

domsj:
Hmm, I would guess that most likely the result would be that
1) either N2 can respond to prepares from N3 or N4 (because N1 shouldn't be flooding N2 with prepares, so in reasonable time prepares from N3 and N4 should get responses) and one of these can get master or
2) N2 itself notices there is no more master and it starts sending prepares, eventually resulting in N2 becoming the master.
In general however there always is the FLP impossibility result, making progress can't be guaranteed.

I do think it would be a good idea to give other nodes a fighting chance by dropping duplicate prepares from the same potential master while already respecting it, so that it never takes arbitrarily long to respond to prepares from other nodes.

toolslive:
respect!

PR #152, opened 2013-08-13
fix inject LeaseExpired event in master_consensus

domsj:
this fixes the failing test_expect_progress_fixed_master in the right suite on CI


domsj:
applied this commit to another branch for which I will soon open a PR

PR #154, opened 2013-08-14
Fix test expect progress fixed master

domsj:
Still waiting for CI, but as usual I'm optimistic and not expecting any problems.

The commits might seem unrelated but together they (the first 2) fix the failing test_expect_progress_fixed_master test.

Will post an update when CI has finished.


domsj:
Unfortunately the initial optimism was uncalled for, however I have now added some commits which do give the desired results on CI.

toolslive:
you probably want gettimeofday iso time there.
otherwise, LGTM.

PR #139, opened 2013-07-30
Cancel fsm alternative

domsj:
Still waiting for CI

Running the slow suites twice to see if any of them hang in catchup while they should be shutting down (such as was the case at http://cislave11.incubaid.com/jenkins/jobs/jenkins-arakoon-git-system-tests-slow-SHAKY-78/home/jenkinslxc/qbase3/var/tmp/arakoon_system_tests/test_sso_deployment/sturdy_2/log/ ).

Will post a comment when CI has finished.

PR #155, opened 2013-08-16
Cancel fsm

domsj:
This PR is largely the same as #139 , but removes some commits that I sneaked into other branches.
The previous PR was also made before the indent changes, which would've made merging a lot harder.
I hereby guarantee that besides the remarks above the changes are the same as in the other.


domsj:
@toolslive can I transfer your LGTM from the other PR to this branch?

toolslive:
yup.

PR #156, opened 2013-08-16
prevent mismatch between number of updates and update_results returned f...

domsj:
...rom store when inserting a value


src/node/store.ml
```diff
@@ -740,7 +740,12 @@ struct
     Logger.debug_f_ "skipped %i updates" j >>= fun () ->
     _insert_updates store updates' kt >>= fun (urs:update_result list) ->
     _with_transaction store kt (fun tx -> _incr_i store tx) >>= fun () ->
-    Lwt.return urs
+    let prepend_oks n l =
+      let rec inner = function
```
toolslive: let rec inner l = function 
   | 0 -> l 
...

PR #157, opened 2013-08-16
for clarity log should always be the first side effect

domsj:
I was at first slightly confused by a test that died during one of the side effects, while the logging for the state had not yet been performed.
This PR should avoid such confusion in the future.


toolslive:
fine.

PR #159, opened 2013-08-19
more finegrained start of lease expirations (int64 -> float)

domsj:
Still waiting for CI
As usual I'm optimistic :-)

This replaces all usages of Unix.time with Unix.gettimeofday where resolution matters.
Protocol has been kept backwards compatible.


domsj:
Ugh, I keep on being too optimistic.
Still trying to figure out what is wrong with this.

domsj:
replaced by #189

PR #160, opened 2013-08-19
Slave accept previous

domsj:
This branch removes the slave_wait_for_accept state, streamlining all into the slave_steady_state state.
The slave now also responds with Accepted to the previously accepted value, this prevents a cluster from getting stuck in case the master for some reason did not receive the previous Accepted message.

CI agrees with these changes

### DIFF
--- a/src/paxos/multi_paxos_fsm.ml
+++ b/src/paxos/multi_paxos_fsm.ml
@@ -123,20 +123,6 @@ let slave_waiting_for_prepare (type s) constants ( (current_i:Sn.t),(current_n:S
           | Nak(n',(n2, i2)) when i2 = current_i ->
             begin
               Logger.debug_f_ "%s: got %s => we're in sync" constants.me (string_of msg) >>= fun () ->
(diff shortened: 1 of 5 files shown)


src/paxos/multi_paxos_fsm.ml
```diff
@@ -127,8 +124,7 @@ let slave_waiting_for_prepare (type s) constants ( (current_i:Sn.t),(current_n:S
             begin
               Logger.debug_f_ "%s: got %s => we're in sync" constants.me (string_of msg) >>= fun () ->
               (* pick in @ steady state *)
-              let p = constants.get_value i2 in
-              match p with
+(*              match p with
```
toolslive: What's this?

toolslive:
This is desirable, but there's some left over comments?

src/paxos/multi_paxos_fsm.ml
domsj: Those are some leftovers indeed. I remember them now :-)

PR #161, opened 2013-08-19
stop the fsms in the tests so tests are better separated

domsj:
This reduces the git base test output from 71 to 59 MB.
Main benefit is that if/when you have to dig through these logs there won't be the overlap anymore.


domsj:
@toolslive says LGTM

PR #162, opened 2013-08-20
prevent mismatch between number of updates and update_results returned f...

domsj:
...rom store when inserting a value


domsj:
Hmm, pull request asks for merge into wrong branch, so closing.

PR #165, opened 2013-08-21
Slave accept previous higher n

domsj:
Some bugfixes that CI agrees with.
Not all is running as blue as I would like yet, still looking into that. But these are good changes, so I would like them to go in.


toolslive:
LGTM.

PR #166, opened 2013-08-21
prevent test_compression_bug and test_compression_previous from failing ...

domsj:
...with longer sleeps

I've seen these fail a few times on CI.
I think it is probably caused by the compression still being busy, so I increased the sleeps.
A better solution would probably involve waiting for the compression to stop, but I don't want to go there.


domsj:
Sigh

PR #167, opened 2013-08-21
correctly obey accepts from other nodes even when in master state

domsj:
CI seems to agree


toolslive:
owkee..

domsj:
That doesn't sound very convinced. If you have doubs or questions just ask, I can be friendly if required.

PR #169, opened 2013-08-22
cleanup election_suggest and read_only state

domsj:
CI looks good


toolslive:
So we actually have a shrinking context here. LGTM

PR #171, opened 2013-08-22
swallow canceled exception from fsm thread

domsj:
CI looks good.
I'm actually not sure anymore if the branch name is really well chosen (I know it dumps a crash log in this case, I actually don't think it hangs).
Should prevent the first 4 debug logs at http://10.101.186.2/jenkins/jobs/jenkins-arakoon-git-system-tests-slow-SHAKY-135/home/jenkinslxc/qbase3/var/tmp/arakoon_system_tests/test_restart_master_long/sturdy_2/log/ from being generated.


toolslive:
LGTM

PR #173, opened 2013-08-23
enable all tests in right suite

domsj:
These give a bloody result on CI, but merging that in is exactly the point.
It should provide the impetus to either remove those tests we don't want anymore or fix those which we still think are relevant.


toolslive:
agree

PR #174, opened 2013-08-23
each node should have its own range of n

domsj:
CI agrees.
If you think you can write it shorter then feel free to do so :-P.


toolslive:
LGTM

PR #178, opened 2013-08-27
adding lcnum and ncnum parameters do config

toolslive:
we might want to make these optional with defaults ?(lcnum=...) But I made them required to get compiler support while doing this.


src/node/local_store.ml
```diff
@@ -80,15 +80,16 @@ let copy_store2 old_location new_location overwrite =
       end
     end
 
-let safe_create db_path mode =
-  Camltc.Hotc.create db_path ~mode [B.BDBTLARGE] >>= fun db ->
+let safe_create ?(lcnum=1024) ?(ncnum=512) db_path ~mode  =
+  (*let lcnum = 8192
+  and ncnum = 4096
+  in*)
```
NicolasT: Remove?

src/node/node_cfg.ml
```diff
@@ -28,7 +28,8 @@ let default_lease_period = 10
 let default_max_value_size = 8 * 1024 * 1024
 let default_max_buffer_size = 32 * 1024 * 1024
 let default_client_buffer_capacity = 32
-
+let default_lcnum = 16384
+let default_ncnum = 8192
```
NicolasT, line 32: Might be strange to have different defaults than the defaults for the optional arguments of `safe_create`

NicolasT:
Some nitpicks, but overall LGTM

src/node/local_store.ml
toolslive: indeed.

src/node/node_cfg.ml
toolslive, line 32: well, sometimes we have different context, I tend to see the defaults in Node_cfg as the ones used in the context of an arakoon node. Sometimes you want to safe_create a store in a different context, with different defaults...

PR #180, opened 2013-09-02
Arakoon 434

toolslive:


NicolasT:
Shouldn't this install a new _camltc_ version in the Jenkins scripts?

NicolasT:
LGTM (if CI agrees & as such camltc changes are OK)

PR #181, opened 2013-09-04
TLS support for Python client

NicolasT:
See: ARAKOON-433
See: http://jira.incubaid.com/browse/ARAKOON-433

### DIFF
--- a/src/client/python/ArakoonClientConnection.py
+++ b/src/client/python/ArakoonClientConnection.py
@@ -20,13 +20,14 @@ GNU Affero General Public License along with this program (file "COPYING").
 If not, see <http://www.gnu.org/licenses/>.
 """
 
+import ssl
 import socket
 from ArakoonProtocol import *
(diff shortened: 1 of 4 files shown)


src/client/python/ArakoonClientConnection.py
```diff
@@ -20,13 +20,14 @@
 If not, see <http://www.gnu.org/licenses/>.
 """
 
+import ssl
```
toolslive, line 23: Is this always available?

toolslive:
otherwise LGTM

src/client/python/ArakoonClientConnection.py
NicolasT, line 23: `ssl` is part of the Python standard library, and should as such be shipped with all somewhat-standard distributions.

If this ever becomes an issue, we can work around it, but for now I'd rather not bloat the code with useless conditionals.

PR #183, opened 2013-09-06
Pylabs: Add TLS configuration methods

NicolasT:
CI passed for `arakoon_system_tests.server.right.test_extension`.

See: ARAKOON-432
See: http://jira.incubaid.com/browse/ARAKOON-432

### DIFF
--- a/extension/server/ArakoonManagement.py
+++ b/extension/server/ArakoonManagement.py
@@ -583,6 +583,199 @@ class ArakoonCluster:
         self._changeFsync(nodes, 'false')
 
 
+    def setTLSCACertificate(self, ca_cert_path):
+        '''Configure path to TLS CA certificate
+
+        This corresponds to the `tls_ca_cert` entry in the `global` section.
+
+        Set to `None` to unset/disable.
+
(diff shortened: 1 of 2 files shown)


extension/server/ArakoonManagement.py
```diff
@@ -583,6 +583,199 @@ def disableFsync(self, nodes=None):
...
+        '''Configure path to TLS CA certificate
+
+        This corresponds to the `tls_ca_cert` entry in the `global` section.
+
+        Set to `None` to unset/disable.
+
+        The path should point to a valid file, otherwise a `ValueError` will be
+        raised.
+
+        :param ca_cert_path: Path to CA certificate
```
domsj, line 596: in other functions I've seen @param instead of :param here; is there any difference?

domsj:
as far as I can tell this looks ok to me

extension/server/ArakoonManagement.py
NicolasT, line 596: Not really a big difference. Maybe things should be consistent, but I noticed several other docstrings have invalid formatting anyway, and we're not actually rendering any docs ever (I'd be surprised if the output would be any good).

PR #184, opened 2013-09-13
Ocaml 4.01 ready

toolslive:
changes to the codebase needed for OCaml 4.01;
changes to our jenkins jobs


domsj:
LGTM

NicolasT:
I'm not very fond of this. Code changes should be OK as long as 4.00 compatibility is kept, but I wouldn't switch to 4.01 "because it's there". Others who build Arakoon will most likely stick to 4.00 for a while (there's no pressing reason to switch to 4.01, as long as 4.00 builds seem stable & we don't use 4.01 features), so our tests should most likely stick to that as well...

toolslive:
maybe we want compiler version as a parameter in jenkins jobs...

PR #187, opened 2013-09-17
Fix 1.6 ci build

domsj:


domsj:
Despite my effort to fix the 1.6 build on travis too in this PR for some reason the travis build still failed, but not with the same error as it otherwise did. I believe my change is correct though.

Edit: opening the PR seems to have triggered building on travis again, let's see if it passes now...

Edit: hah, all is green now, great.

PR #190, opened 2013-09-17
System tests ssl

toolslive:
This is not finished, but this is way better than failing the whole suite.


domsj:
Ok, is good for me.
Regarding config of arakoon with tls stuff: Nicolas still has an open pr #183 which you might want to use.

PR #201, opened 2013-09-30
ensure lease expirations never fire too early

domsj:
I noticed in logs from CI that a certain LeaseExpiration was delivered too early, indicating Lwt_unix.sleep can return too early.
( Sep 27 15:31:11 7028: (paxos|debug): sturdy_1: lease expired (8.0 passed, intended 8.0, diff -0.000040)=> injecting LeaseExpired event for 11 )
This PR should fix that problem.


toolslive:
really odd that this goes against what libev 'guarantees'

domsj:
maybe it's related to how the gettimeofday call and how it provides sub second resolution?
and/or the process being continued on another core?
Either way, I've seen it happen ...

PR #206, opened 2013-10-04
2MB iso 8

toolslive:


domsj:
If I'm correct this doesn't change reading of (old) tlogs.
Could it be useful to keep the magic 900kB in mind that is used as the bz2 block size?
Otherwise, LGTM.

PR #208, opened 2013-10-07
implement --force option for --inject-as-head to allow replacing a curre...

domsj:
...nt corrupt head.db

add different exit codes depending on which database is corrupted

### DIFF
--- a/src/node/dump_store.ml
+++ b/src/node/dump_store.ml
@@ -46,7 +46,9 @@ let dump_store filename =
   Lwt_main.run (t());
   0
 
-let inject_as_head fn node_id cfg_fn =
+exception ExitWithCode of int;;
+
+let inject_as_head fn node_id cfg_fn force =
   let canonical =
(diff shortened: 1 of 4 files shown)


domsj:
FYI this is related to http://jira.incubaid.com/browse/ARAKOON-438
CI agrees with these changes

src/node/dump_store.ml
```diff
@@ -46,7 +46,9 @@ let dump_store filename =
   Lwt_main.run (t());
   0
 
-let inject_as_head fn node_id cfg_fn =
+exception ExitWithCode of int;;
```
toolslive, line 49: ;; ?

toolslive:
LGTM

src/node/dump_store.ml
domsj, line 49: That's not needed indeed, not really sure why I put that there.
I'll just merge it in as is though...

PR #209, opened 2013-10-07
ARAKOON-418 fsync=true should be the default, warning otherwise

domsj:


domsj:
Hmm, maybe change python integration tests to set fsync=false, as we know what we're doing?

domsj:
This finally looks good on CI

toolslive:
LGTM

PR #211, opened 2013-10-07
Use a PPA iso building OCaml & opam from source

NicolasT:
These commits configure TravisCI to use PPAs provided by @avsm containing different OCaml & opam versions. This speeds up our build phase, and as a result 'arakoon.native --run-all-tests' fits into a TravisCI job time-slot as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,18 @@
+language: c
+env:
+  - PPA=avsm/ocaml40+opam10
+  - PPA=avsm/ocaml40+opam11
+  - PPA=avsm/ocaml41+opam10
+  - PPA=avsm/ocaml41+opam11
+
 before_install:
+  - echo "yes" | sudo add-apt-repository ppa:$PPA
   - sudo apt-get update -qq
-  - sudo apt-get install -qq ocaml libev-dev libssl-dev
+  - sudo apt-get install -qq ocaml ocaml-native-compilers camlp4-extra opam libev-dev libssl-dev
+  - ocamlopt -version
 
 install:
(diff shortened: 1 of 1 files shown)


domsj:
LGTM

avsm:
Let me know if you need any more PPA variants for your tests.  I'm working on an ARM recipe that caches the intermediate chroot so it fits within the 50 minute budget.

-anil

domsj:
Hey @avsm , your ocaml40+opam10 environment seems to be giving segfaults on travis when doing "opam init -v -y".
All the other envs work fine. Do you know what could be wrong there? Is it something that is fixed in the latest opam10 release which has not yet made it to your image?
See https://travis-ci.org/Incubaid/arakoon/jobs/12265285 for an example.

Thanks in advance!
Jan

avsm:
This is indeed an upstream bug in the older versions (that only happens on 3.12.1) -- still tracking it down.

domsj:
Ok, thanks for the quick response.
It's not just ocaml 3.12.1 but also 4.00.1 right? (As that's the ocaml version in your image)

avsm:
Hm, we're only seeing this in 3.12.1. I'll have to try and repro locally.  I suspect it's just a stack overflow.

domsj:
@avsm our travis ci build (https://travis-ci.org/Incubaid/arakoon/builds 1104 -> 1007) is succeeding again for the PPA=avsm/ocaml40+opam10 environment, despite us not changing anything... so not really sure what's going on.
Thought you might want to know :-).

Jan

PR #217, opened 2013-10-14
Copy store with temp file

domsj:
By tackling the file copy through a temp file which gets renamed at the Filesystem.copy_file level I'm quite sure I covered all situations we want to cover.

### DIFF
--- a/src/tools/file_system.ml
+++ b/src/tools/file_system.ml
@@ -24,30 +24,6 @@ open Lwt
 
 let section = Logger.Section.main
 
-let copy_file source target = (* LOOKS LIKE Clone.copy_stream ... *)
-  Logger.info_f_ "copy_file %s %s" source target >>= fun () ->
-  let bs = Lwt_io.default_buffer_size () in
-  let buffer = String.create bs in
-  let copy_all ic oc =
-    let rec loop () =
-      Lwt_io.read_into ic buffer 0 bs >>= fun bytes_read ->
-      if bytes_read > 0
-      then
-        begin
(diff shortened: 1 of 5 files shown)


src/tools/file_system.ml
```diff
@@ -71,10 +47,16 @@ let lwt_directory_list dn =
     )
     (fun exn -> Logger.debug_f_ ~exn "lwt_directory_list %s" dn >>= fun () -> Lwt.fail exn)
 
+let fsync_dir filename =
```
NicolasT: This might raise false expectations with other users of this procedure. When I see something like

``` ocaml
val fsync_dir : string -> unit Lwt.t
```

I expect this to fsync the path of the directory I pass to the action. This code will fsync the parent directory instead.

How about keeping `fsync_dir` as is, but without the call to `Filename.dirname` (and renaming the argument) and moving `Filename.dirname` to the call site (or figure out some good name for an action which _does_ expect a file path argument & performs the `dirname` call as required)?

domsj: fsync_dir_of_file ?

NicolasT:
This got me thinking: when creating/opening a new `tlog` file, shouldn't the corresponding dirent be synced as well? Otherwise we might start writing to the tlog (maybe even using `fsync`), but if a hardware crash occurs at the wrong time, the file is gone anyway.

domsj:
That's a good point, and indeed a situation that is not yet covered here.
I'll add another commit for that.

NicolasT:
If CI is OK, LGTM

PR #218, opened 2013-10-16
close of store and tlog shouldn't happen when an exception occured

domsj:


domsj:
Cucu! I've been waiting here for a week.

toolslive:
ok.

PR #222, opened 2013-10-18
sane defaults for batched store

toolslive:
same defaults in example cfg
compression on 2MB iso 8MB
log which tokyo cabinet is slow


domsj:
LGTM

PR #225, opened 2013-10-22
ensure tlogcollection has enough time to create the desired compression ...

domsj:
...jobs before already closing it


domsj:
Some explanation: I saw this test fail once, and I think the reason is that the tlog_collection did not yet have time to create these compression tasks.
I'll admit the solution is 'turkish' .. but it should probably be good enough to never see the error again.
If I do see it again I'm willing to think about doing some synchronizing to solve it ...

toolslive:
ok

PR #227, opened 2013-10-23
Enable more tests

domsj:
These tests fail about half of the time (tested locally and on CI, http://cimaster.incubaid.com/view/arakoon-git/job/arakoon-git-base/528/testReport/junit/universe.system/startup/restart_slaves/ ).
When merged into 1.6 head they should however succeed again (already tested locally).


domsj:
I suppose the 'yeah!' at #228 is also applicable here :).

PR #229, opened 2013-10-24
Backport forced master

domsj:
This backports a fix to 1.6 for the situation where a forced master too quickly starts responding to requests from clients.
The first 2 commits are cherry-picked from the 1.7 branch.
I think we should actually still discuss if we really want to backport this ...


domsj:
@toolslive said it's ok to merge this one.

PR #230, opened 2013-10-25
drop connection when losing master role while trying to push an update

domsj:
(to indicate to the client the update might or might not have succeeded)

Also check the other pull request with an alternative implementation.

### DIFF
--- a/src/node/store.ml
+++ b/src/node/store.ml
@@ -67,6 +67,7 @@ class transaction = object end
 
 type update_result =
   | Ok of string option
+  | Lost_master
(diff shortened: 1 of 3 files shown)


NicolasT:
No need for this. If this gets merged in 1.6, it will be forward-ported when 1.6 is merged into 1.7.

domsj:
The other pull request is actually against 1.7 too, but the implementation is slightly different.
So you and/or Romain would have to choose which variation you like better...
I suggest choosing first and then I can backport this change against 1.6 in a pull request.

domsj:
will be replaced with another PR

PR #231, opened 2013-10-25
drop connection when losing master role while trying to push an update (...

domsj:
...to indicate to the client the update might or might not have succeeded)

Also see #230


toolslive:
fine.

domsj:
will be replaced with another PR

PR #232, opened 2013-10-25
change batched_store to use only 1 internal hashtbl (simpler and more ef...

domsj:
...ficient)

But before we merge this in we probably want to do some performance testing...


domsj:
Ok, I did a small test, with 18 tlogs/tlfs generated by --benchmark.
Regular 1.7: 2min21s
This branch: 2min12s

Small difference, might be noise.
I still think this change is useful though...

PR #234, opened 2013-10-31
Arakoon 445

toolslive:


src/node/catchup.ml
```diff
@@ -267,7 +267,7 @@ let catchup_store ~stop (type s) me
               | None ->
                 let n = tlog_coll # get_tlog_from_i ssi in
                 let ni = Sn.to_int n in
-                let an = Tlc2.archive_name ni
+                let an = Tlc2.archive_name Compression.default ni
```
domsj: shouldn't it look for both .tls and .tlf when it can't find the tlog? so the log message being built here should probably reflect that.

src/node/node_cfg.ml
```diff
@@ -413,7 +415,21 @@ module Node_cfg = struct
...
+      then Compression.No
+      else
+        begin
+          let s = Ini.get inifile node_name "tlog_compression" Ini.p_string
+                          (Ini.default "bz2")
+          in
+          match String.lowercase s with
+          | "snappy"  -> Compression.Snappy
+          | "bz2"     -> Compression.Bz2
+          | "none"    -> Compression.No
```
domsj, line 429: Too lazy to look for the right spot, but I would expect Compression.None here iso x.No

src/tlog/compression.ml
```diff
@@ -23,15 +23,32 @@ If not, see <http://www.gnu.org/licenses/>.
...
 open Tlogcommon
 
-let archive_name tlog_name = tlog_name ^ ".aar"
-let tlog_name archive_name =
-  let len = String.length archive_name in
-  let ext = String.sub archive_name (len-4) 4 in
-  assert (ext=".aar");
-  String.sub archive_name 0 (len-4)
+type compressor =
+  | No
```
domsj, line 27: Found the spot ;-)

src/tlog/tlc2.ml
```diff
@@ -32,15 +32,28 @@ exception TLCNotProperlyClosed of string
...
 let section = Logger.Section.main
 
-let file_regexp = Str.regexp "^[0-9]+\\.tl\\(og\\|f\\|c\\)$"
+let file_regexp = Str.regexp "^[0-9]+\\.tl\\(og\\|f\\|c\\|s\\)$"
+
 let tlog_regexp = Str.regexp "^[0-9]+\\.tlog$"
 let file_name c = Printf.sprintf "%03i.tlog" c
-let archive_extension = ".tlf"
-let archive_name c = Printf.sprintf "%03i.tlf" c
+let archive_extension = ".tls"
```
domsj: Why is .tls more special than .tlf ?
This will probably lead to bugs where code only works against .tls and not against .tlf.

src/tlog/tlc2.ml
```diff
@@ -163,7 +178,7 @@ let fold_read tlog_dir tlf_dir file_name
             Lwt_io.with_file ~mode:Lwt_io.input full_an
               (fun ic ->
                  let index = None in
-                 Tlogreader2.AC.fold ic ~index lowerI too_far_i ~first a0 f)
+                 Tlogreader2.AS.fold ic ~index lowerI too_far_i ~first a0 f)
```
domsj: This part of the code only does fallback to .tls

src/tlog/tlc2.ml
```diff
@@ -545,19 +563,20 @@ class tlc2 (tlog_dir:string) (tlf_dir:string) (head_dir:string) (new_c:int)
...
+           let tlu = get_full_path tlog_dir tlf_dir (file_name old_outer) in
+           let tlc = get_full_path tlog_dir tlf_dir
+                                   (archive_name compressor old_outer) in
+           let tlc_temp = tlc ^ ".part" in
+           let job = (tlu,tlc_temp,tlc,compressor) in
+           Logger.info_f_ "adding new task to compression queue %s,%s,%s"
+                          tlu tlc_temp tlc >>= fun () ->
+           Lwt_buffer.add job _compression_q
+         end
+      | No -> Lwt.return ()
```
domsj, line 603: Should this case move the tlog to the tlx_dir ?

src/tlog/tlc2.ml
```diff
@@ -721,34 +740,36 @@ class tlc2 (tlog_dir:string) (tlf_dir:string) (head_dir:string) (new_c:int)
...
-        File_system.exists (get_full_path tlog_dir tlf_dir an) >>= function
-        | true -> Lwt.return an
-        | false -> Lwt.return fn
-      end
-      >>= fun name ->
-      let canonical = get_full_path tlog_dir tlf_dir name in
+      self # which_tlog_file start_i >>= fun co ->
+      let canonical = match co with
+        | Some fn -> fn
+        | None -> let n = Sn.to_int (get_file_number start_i) in
```
domsj, line 791: Log a warning an throw a meaningful exception here?
The file name being constructed in this code path will not refer to an existing file anyway (according to the logic in which_tlog_file)

src/tlog/tlc2_test.ml
```diff
@@ -32,7 +32,8 @@ let section = Logger.Section.main
 
 let create_test_tlc dn =
   let tlf_dir = (dn ^ "_tlf") in
-  Tlc2.make_tlc2 dn tlf_dir tlf_dir true false
+  let compressor = Compression.default in
```
domsj: Run tests for each type of compressor instead of hardcoding it the default one?
Using Compression.default should (most of the time) by considered wrong.

src/tlog/tlc2_test.ml
```diff
@@ -252,9 +253,10 @@ let test_compression_previous (dn, tlf_dir, factory) =
   tlc # close ~wait_for_compression:true () >>= fun () ->
 
   (* mess around : uncompress tlfs to tlogs again, put some temp files in the way *)
-  let uncompress tlf =
-    let tlfpath =  (tlf_dir ^ "/" ^ tlf ^ ".tlf") in
-    Compression.uncompress_tlog tlfpath (dn ^ "/" ^ tlf ^ ".tlog") >>= fun () ->
+  let ext = Tlc2.archive_extension in
```
domsj: Should be dependent on the compressor being used in the test. (I know, right now it only tests snappy, so for now it was correct.)

src/tlog/tlc2_test.ml
```diff
@@ -264,14 +266,14 @@ let test_compression_previous (dn, tlf_dir, factory) =
 
   let touch fn =
     let fd = Unix.openfile fn [Unix.O_CREAT] 0o644 in Unix.close fd in
-  touch (tlf_dir ^ "/000.tlf.part");
+  touch (tlf_dir ^ "/000" ^ ext ^ "s.part");
```
domsj: `touch (tlf_dir ^ "/000" ^ ext ^ ".part");`

src/tlog/tlog_main.ml
```diff
@@ -108,12 +108,13 @@ let compress_tlog tlu =
...
-  let tlf = Tlc2.to_archive_name tlu in
-  let () = if Sys.file_exists tlf then failwith "Can't compress %s as %s already exists" tlu tlf in
+  let tlx = Tlc2.to_archive_name tlu in
+  let () = if Sys.file_exists tlx then failwith "Can't compress %s as %s already exists" tlu tlx in
   let t =
-    let tmp = tlf ^ ".tmp" in
-    Compression.compress_tlog tlu tmp >>= fun () ->
-    File_system.rename tmp tlf >>= fun () ->
+    let tmp = tlx ^ ".tmp" in
+    Compression.compress_tlog ~cancel:(ref false) tlu tmp Compression.default
```
domsj: Probably it should be possible to specify which compressor should be used from the command line.

extension/server/ArakoonManagement.py
```diff
@@ -533,34 +533,35 @@ def setLogLevel(self, level, nodes=None):
...
+        self._setTlogCompression(nodes,compressor)
+
     def disableTlogCompression(self, nodes=None):
         """
         Disables tlog compression for the given nodes
         @param nodes List of node names
         """
-        self._changeTlogCompression(nodes, 'true')
-
+        self._setTlogCompression(nodes,"none")
```
domsj, line 564: Ok, so this doesn't set the `disable_tlog_compression` flag anymore for backwards compatibility. (Fine by me.)
Maybe now is also the time to think about what happens with old tlogs that don't have to be compressed?
Also, should we rename tlf_dir to tlx_dir?

extension/test/server/left/system_tests_long_left.py
```diff
@@ -164,3 +164,49 @@ def load():
...
+    tlx_dir = config.get('tlf_dir')
+    if not tlx_dir:
+        tlx_dir = config.get('home')
+    files = os.listdir(tlx_dir)
+
+    tls = filter(lambda x:x.endswith(".tls"), files)
+    tlf = filter(lambda x:x.endswith(".tlf"), files)
+    assert_true(len(tls) > 0, "we should have .tls files" )
+    assert_true(len(tlf) > 0, "we should have .tlf files" )
+    # does catchup still work?
```
domsj, line 197: So this test verifies that node n0 has both .tls and .tlf; and after that it is wiped.
I would've expected n1 to be wiped at this point, so that n1 can do catchup from a node that has both .tls and .tlf files.

src/tlog/compression.ml
```diff
@@ -23,15 +23,32 @@ If not, see <http://www.gnu.org/licenses/>.
...
+  | Bz2
+  | Snappy
+
+let compressor2s = function
+  | No -> "No"
+  | Bz2 -> "Bz2"
+  | Snappy -> "Snappy"
+
+let default = Snappy
```
domsj: Isn't it a little strange to have Bz2 be the default according to the config and have it as Snappy here?

domsj:
I think in the (python) tests we use the default compression.
Do we want to change most of the tests to use snappy instead of bz2?
Could that be better for the load on our CI slaves?
Just throwing the question out there so we can make a conscious decision about it.

src/tlog/tlc2.ml
toolslive, line 603: Yes, we could streamline this; and fix the problem that is we disable compression the files don't end up in the 
"_tlf" dir. the name tlf_dir is also increasingly wrong.

src/tlog/tlc2.ml
toolslive, line 791: there's a general function to find the canonical filename (%s/"%03i.%s"  % (path,counter, extension), given the counter i. we should use it everywhere.

src/tlog/tlc2.ml
domsj: When tlogs get moved to the tlx_dir (not sure if you're intent on implementing that in this PR) then there should also be a fallback to the moved tlog here.
Maybe 'compressor' should more be considered a tlog post processor with options bz2, snappy(, none) and move?

domsj:
LGTM

PR #235, opened 2013-11-04
in filter tls < tlf < tlog; extension cleanup

toolslive:


domsj:
streamline this priority with the code in `find_tlog_file` ?

src/tlog/compression_test.ml
```diff
@@ -64,14 +64,14 @@ let test_compress_file which () =
...
-  let archive_name x = x ^ ".tls"
+let snappy =
+  let archive_name x = x ^ ".tlx"
   and compressor = Compression.Snappy
   and tlog_name a = Filename.chop_extension a
   in
   (archive_name, compressor, tlog_name)
 
-let snappy =
+let bz2 =
```
domsj, line 74: hehe

src/tlog/compression.ml
```diff
@@ -44,117 +44,138 @@ let uncompress_bz2 b =
 let compress_snappy b = Snappy.compress b
 let uncompress_snappy b = Snappy.uncompress b
 
+
+let format_snappy = "snappy"
+let read_format ic = function
```
domsj, line 49: So this code is used to verify that the right decompressor is being used when reading a file, right?
But shouldn't the format be read to be able to select the right decompressor in the first place?

domsj:
Except for the comment about `read_format` this looks good to me.

src/tlog/compression.ml
toolslive, line 49: putting bz2 inside .tlx too will be left for later.

PR #238, opened 2013-11-05
throw an error when sabotage is detected (instead of unpacking last .tlf...

domsj:
.../.tlx)

### DIFF
--- a/extension/test/server/right/system_tests_long_right.py
+++ b/extension/test/server/right/system_tests_long_right.py
@@ -488,9 +488,7 @@ def test_sabotage():
         print f
         q.system.fs.remove(f)
 
-    cmd = []
-    command = clu._cmd('sturdy_0')
-    cmd.extend(command)
+    cmd = clu._cmd('sturdy_0')
     returncode = subprocess.call(cmd)
     assert_equals(returncode, 50)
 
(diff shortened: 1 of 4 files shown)


extension/test/server/right/system_tests_long_right.py
```diff
@@ -487,18 +487,12 @@ def test_sabotage():
...
-    Common.iterate_n_times(2000, Common.simple_set)
-    logging.info("sleeping fo 10s")
-    time.sleep(10)
 
-    size = q.system.fs.fileSize("%s/001.tlf" % node_tlf_dir)
-    logging.info("file_size = %i", size)
-    assert_true(size > 1024 * 5)
+    cmd = []
+    command = clu._cmd('sturdy_0')
+    cmd.extend(command)
```
toolslive: cmd = clu._cmd('sturdy_0')

src/node/node_main.ml
```diff
@@ -801,6 +801,10 @@ let _main_2 (type s)
             let rc = 49 in
             Logger.fatal_f_ "[rc=%i] Tlog has a checksum error, last valid pos = %Li" rc pos >>= fun () ->
             Lwt.return rc
+          | Tlogcommon.TLogSabotage ->
+            let rc = 50 in
+            Logger.fatal_f_ "[rc=%i] Somebody has been messing with the available tlogs" rc >>= fun () ->
+            Lwt.return rc
```
toolslive, line 807: I just wonder how long it will take before this gets reported back to us ;)

toolslive:
tentative LGTM

src/node/node_main.ml
domsj, line 807: Hehe, I wonder the same thing. But it should definitely make them think a little about what they're doing before they harass us :-)

PR #240, opened 2013-11-06
log time it took to commit the tc transaction iso of the whole transacti...

domsj:
...on (related to recent batched store changes)


toolslive:
LGTM.

PR #244, opened 2013-11-08
Slowdown collapse

domsj:
Note this is a pull request against 1.6
Should be forward merged/ported later.

### DIFF
--- a/cfg/arakoon.ini
+++ b/cfg/arakoon.ini
@@ -60,7 +60,20 @@ home = /tmp/arakoon/arakoon_0
 log_level = debug
 log_config = default_log_config
 batched_transaction_config = default_batched_transaction_config
-#collapse_slowdown = 2
+
+# Throttle transaction processing during collapsing
+# On some hardware and in some deployments, the collapsing process can be a
+# very IO and CPU-hungry process. This can have a negative impact on the
+# overall responsiveness of an Arakoon node.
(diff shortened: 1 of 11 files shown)


NicolasT:
This needs more documentation. It's fairly unclear to me (even while looking at the patch) what this actually _does_, or what the impact of this 'factor' would be. What's a sensible value? What happens if I increase/decrease this value?...

domsj:
How do you mean it's unclear what it does?
During the collapse it sleeps for a certain while after each transaction. How long it sleeps depends on the configured factor. Imnsho that should be quite clear if you look at the code.

Probably 1, 2 or 3 are sensible values, but to be sure what value is best this should probably be determined by trying it with a few different values (which I believe Mattias is doing right now). The best value might depend on the IO and cpu capabilities of the system it's running on and the load it receives.

I refuse to answer the question about what happens when the value increases or decreases, it should be trivial to answer it yourself if you look at what the code does.

Feel free to make concrete suggestions about how exactly it should be better documented.

toolslive:
for what it's worth: I perfectly understand what it is supposed to do (and so does our volunteer tester)
LGTM.

domsj:
I'm still waiting for Mattias to report his findings back to us... will merge it in if the result is positive.

domsj:
#308 , which just got merged, handles this on the 1.7 branch. unless specifically request this won't be available on 1.6.

PR #246, opened 2013-11-08
Misc fixes

toolslive:
disabling coverage cuts down the duration to 11min on jenkins


domsj:
What's the plan with the coverage thingie? Run manual from time to time, or make another job for it?
The changes here look OK.

PR #248, opened 2013-11-13
TravisCI: Rework build using a script

NicolasT:
In order to make sure a build actually fails whenever a build-step
fails, this commit moves the steps into a shell-script which is then
executed through the TravisCI configuration file.


NicolasT:
Don't merge yet, this is supposed to fail for now.

NicolasT:
Added dependencies. This one is supposed to work.

domsj:
LGTM

PR #251, opened 2013-11-19
Master state

toolslive:


domsj:
Change looks good.
Do we want this against 1.6 or 1.7?
Probably want to run the CI tests first too.

NicolasT:
Agree with @domsj, this should go into 1.6.6?

PR #253, opened 2013-11-20
Pass 'open modes' instead of 'quiesced' flags to a store

NicolasT:
In line with a comment made in #252, this commit changes some code in
the `Store` subsystem to not accept `bool` values denoting
'quiesced mode' (something a store should be oblivious of), but use
'open mode' flags instead.

### DIFF
--- a/src/node/local_store.ml
+++ b/src/node/local_store.ml
@@ -67,8 +67,16 @@ let copy_store2 old_location new_location overwrite =
   else
     File_system.copy_file old_location new_location overwrite
 
+let bdb_open_mode mode =
+  let map_mode = function
+    | OpenMode.OREADER -> Camltc.Bdb.OpenMode.OREADER
+    | OpenMode.OWRITER -> Camltc.Bdb.OpenMode.OWRITER
+    | OpenMode.OCREAT -> Camltc.Bdb.OpenMode.OCREAT
+  in
+  Camltc.Bdb.OpenMode.OLCKNB :: List.map map_mode mode
+
 let safe_create db_path mode =
-  Camltc.Hotc.create db_path ~mode [B.BDBTLARGE] >>= fun db ->
+  Camltc.Hotc.create db_path ~mode:(bdb_open_mode mode) [B.BDBTLARGE] >>= fun db ->
   let flags = Camltc.Bdb.flags (Camltc.Hotc.get_bdb db) in
   if List.mem Camltc.Bdb.BDBFFATAL flags
     then Lwt.fail (BdbFFatal db_path)
@@ -191,15 +199,8 @@ let close ls flush =
 let get_location ls = Camltc.Hotc.filename ls.db
 
 let reopen ls f mode =
-  let map_mode = function
-    | OpenMode.OREADER -> Camltc.Bdb.OpenMode.OREADER
-    | OpenMode.OWRITER -> Camltc.Bdb.OpenMode.OWRITER
-    | OpenMode.OCREAT -> Camltc.Bdb.OpenMode.OCREAT
-  in
-  let bdb_mode = Camltc.Bdb.OpenMode.OLCKNB :: List.map map_mode mode in
-
   Logger.info_f_ "local_store %S::reopen calling Hotc::reopen" ls.location >>= fun () ->
-  Camltc.Hotc.reopen ls.db f bdb_mode >>= fun () ->
+  Camltc.Hotc.reopen ls.db f (bdb_open_mode mode) >>= fun () ->
   Logger.info_ "local_store::reopen Hotc::reopen succeeded"
 
 let get_key_count ls =
@@ -236,7 +237,7 @@ let optimize ls mode =
   >>= fun () ->
   begin
     Logger.info_f_ "Creating new db object at location %s" db_optimal >>= fun () ->
-    safe_create db_optimal Camltc.Bdb.default_mode >>= fun db_opt ->
+    safe_create db_optimal [Store.OpenMode.OCREAT; Store.OpenMode.OWRITER] >>= fun db_opt ->
     Lwt.finalize
       ( fun () ->
         Logger.info_ "Optimizing db copy" >>= fun () ->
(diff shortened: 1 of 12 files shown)


NicolasT:
This is not ready for merge yet. It should pass CI, but this depends on toolslive/camltc#9.

src/node/store.ml
```diff
@@ -71,6 +71,12 @@ type update_result =
 exception Key_not_found of string
 exception CorruptStore
 
+module OpenMode = struct
+  type t = OREADER
+         | OWRITER
+         | OCREAT
```
domsj, line 77: What's the reason for having OCREAT?
(And if there is none, why not just keep it a boolean?)

src/node/local_store.ml
```diff
@@ -190,16 +190,16 @@ let close ls flush =
...
-      else
-        B.default_mode
-    end in
+let reopen ls f mode =
+  let map_mode = function
+    | OpenMode.OREADER -> Camltc.Bdb.OpenMode.OREADER
+    | OpenMode.OWRITER -> Camltc.Bdb.OpenMode.OWRITER
+    | OpenMode.OCREAT -> Camltc.Bdb.OpenMode.OCREAT
+  in
+  let bdb_mode = Camltc.Bdb.OpenMode.OLCKNB :: List.map map_mode mode in
```
domsj: This results in other modes for opening the store then it was before, is that intentional?
I would've kept on using `B.readonly_mode` and `B.default_mode`, that's why they exist in the first place, no?

src/node/store.ml
NicolasT, line 77: Because in some circumstances we want to open an _existing_ database for access (R or W), especially when using `reopen`. The prior behavior using `Bdb.default_flags` added the `BDBOCREAT` flag, which is (I think) buggy.

src/node/local_store.ml
NicolasT: In places where `Bdb.default_flags` was used before with the intention of creating a database, this is still kept as-is. So yes, in some other places (`reopen`), this is a (I think correct) change in behavior.

IMHO having something like `Bdb.default_flags` is a somewhat broken API.

src/node/store.ml
domsj, line 77: Agreed that when reopening you don't want to pass in the OCREAT flag.
Before, when opening it for writing it also added the OREADER flag, is that not needed here? (Maybe writing implies reading, but then definition of default_mode in camltc is a bit redundant.)
Your answer doesn't give a reason for having OCREAT though. Actually you just argued that having this option available and passed into the reopen function is buggy.

NicolasT:
This last commit, a99a845, must certainly pass CI in order to make sure correct flags are passed at all times (read/write mode & creation).

src/node/store.ml
NicolasT, line 77: `OCREAT` makes sense in code paths where a database is supposed to be created.

a99a845 removes the use of `Bdb.default_mode` and makes all open modes explicit.

src/node/store.ml
```diff
@@ -88,15 +94,15 @@ module type Simple_store = sig
 
   val flush: t -> unit Lwt.t
   val close: t -> bool -> unit Lwt.t
-  val reopen: t -> (unit -> unit Lwt.t) -> bool -> unit Lwt.t
-  val make_store: bool -> string -> t Lwt.t
+  val reopen: t -> (unit -> unit Lwt.t) -> OpenMode.t list -> unit Lwt.t
```
domsj, line 97: I still think this function should only allow passing in 2 values, designating either opening it read-only or read-write.
(A good choice to represent 2 values in a lot of cases is a boolean, perhaps as a labeled argument if that suits you better.)

NicolasT, line 97: Disagree. `t` is some opaque handle to some underlying database. What if I want to `reopen` the database underlying this handle using `[OCREAT; OWRITER]` after I unlinked the underlying database (for some reason)?

domsj, line 97: When would you want to do that? As far as I know we never want to do that in arakoon, so why have the option in the API?
Unused code paths and possibilities are imo technical debt.

NicolasT, line 97: Feel free to change according to your insights, opinions and needs.

src/node/store.ml
```diff
@@ -71,6 +71,12 @@ type update_result =
 exception Key_not_found of string
 exception CorruptStore
 
+module OpenMode = struct
```
toolslive, line 74: why create a module just for a type, and then prefix the constructor with 'O' ?
(I'm not even sure the type is necessary)

src/node/store.ml
```diff
@@ -113,7 +119,7 @@ module type STORE =
   sig
     type ss
     type t
-    val make_store : ?read_only:bool -> string -> t Lwt.t
+    val make_store : OpenMode.t list -> string -> t Lwt.t
```
toolslive, line 122: This creates 'ballast'

src/node/store_test.ml
```diff
@@ -108,11 +108,11 @@ let test_safe_insert_value () =
           asserts)
       value_asserts in
   Logger.info_ "applying updates without surrounding transaction" >>= fun () ->
-  with_store "tsiv" (fun store ->
+  with_store [Store.OpenMode.OCREAT; Store.OpenMode.OWRITER] "tsiv" (fun store ->
```
toolslive, line 111: This is exactly the `ballast` I don't want in tests.
(btw, what's wrong with the 'open' keyword ?)

src/paxos/multi_paxos_test.ml
```diff
@@ -65,7 +65,7 @@ let test_generic network_factory n_nodes () =
   let last_witnessed who = Sn.of_int (-1000) in
   let inject_buffer = Lwt_buffer.create_fixed_capacity 1 in
   let inject_ev q e = Lwt_buffer.add e q in
-  S.make_store "MEM#store" >>= fun store ->
+  S.make_store [Store.OpenMode.OCREAT; Store.OpenMode.OWRITER] "MEM#store" >>= fun store ->
```
toolslive, line 68: for a memstore/test context I don't care, and I would like not to be bothered with these modes.

PR #255, opened 2013-11-25
Rework TLS tests

NicolasT:
The TLS tests should now cover what was tested using a shell-script before.

This passes `arakoon-git-test-spec` using `arakoon_system_tests.server.quick.system_tests_ssl` (see job 30).


toolslive:
fine.

PR #260, opened 2013-12-03
sprinkle some time.sleep(1.0) in the tests

domsj:


extension/test/server/quick/system_tests_basic.py
```diff
@@ -336,6 +337,7 @@ def test_test_and_set() :
 
 @C.with_custom_setup( C.setup_3_nodes_forced_master , C.basic_teardown )
 def test_who_master_fixed () :
+    time.sleep(1.0)
```
NicolasT: Wouldn't it make sense to put this sleep in `C.setup_3_nodes_forced_master` (and `C.setup_3_nodes` and maybe others)?

NicolasT:
OK assuming this actually fixes/enhances things.

domsj:
I think it does. If future CI test runs show it doesn't we should start using those wait_for_master helpers you pointed to on the chat.

PR #262, opened 2013-12-04
sleep a bit after starting a node and before asserting how many nodes ar...

domsj:
...e running

based on

http://172.19.49.85/view/arakoon-git/job/arakoon-amplidata-git-system-tests-slow-RIGHT/lastCompletedBuild/testReport/arakoon_system_tests.server.right/system_tests_learner/test_learner/
http://10.101.186.8/jenkins/jobs/jenkins-arakoon-amplidata-git-system-tests-slow-RIGHT-9/home/jenkinslxc/qbase3/var/tmp/arakoon_system_tests/test_learner/sturdy_2/log/

### DIFF
--- a/extension/test/server/right/system_tests_learner.py
+++ b/extension/test/server/right/system_tests_learner.py
@@ -49,6 +49,8 @@ def test_learner():
     cluster.createDirs(name)
     cluster.startOne(name)
 
+    time.sleep(1.0)
+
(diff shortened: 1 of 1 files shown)


NicolasT:
If this enhances test success rates, LGTM. We should find a more 'correct' way instead of putting random sleeps all over the tests though.

PR #266, opened 2013-12-10
Lost master client exception

domsj:
looks good on CI


src/client/client_protocol.ml
```diff
@@ -89,10 +89,12 @@ let handle_exception oc exn=
     match exn with
       | XException(Arakoon_exc.E_NOT_FOUND, msg) ->
         Lwt.return (Arakoon_exc.E_NOT_FOUND,msg, false, false, Logger.Debug)
-      | XException(Arakoon_exc.E_GOING_DOWN, msg) ->
```
NicolasT, line 92: Any reason to move this around?

domsj, line 92: It looked nicer to put the XExceptions with an error code resulting `..., false, false, Logger.Debug` together.

toolslive:
fine.

PR #267, opened 2013-12-11
Fix masterless cluster

domsj:


toolslive:
fine, on the condition CI passes.

PR #268, opened 2013-12-11
Return `E_NOT_SUPPORTED` from slave-only calls on master

NicolasT:
This commit changes the error returned from `_not_if_master` and `quiesce_db` from `E_UNKNOWN_FAILURE` (which closes the client socket) to `E_NOT_SUPPORTED` (which keeps the client socket as-is) in case the call is only valid on non-master nodes.


NicolasT:
Running CI: `arakoon-git-launcher` 163.

domsj:
Looks good if CI agrees.

PR #270, opened 2013-12-11
update Makefile install destination dirs

domsj:


NicolasT:
Maybe slightly alter the commit message to point to ocaml/opam-repository#1423?

domsj:
Sure, done

PR #277, opened 2013-12-16
fix test_download_db

domsj:
This should probably fix the failure I saw at http://172.19.49.85/view/arakoon-git/job/arakoon-amplidata-git-system-tests-QUICK/31/testReport/junit/arakoon_system_tests.server.quick/system_tests_basic/test_download_db/ .
If you have better ideas, which are applicable here, then please share.


toolslive:
I guess the main problem is ArakoonClientConnection.py. It tries to connect once, and then only once for each message to send. Since we want to move this client out, I would not bother.
LGTM.

PR #279, opened 2013-12-17
don't go to election while still respecting another node running for mas...

domsj:
...ter

Looks good on CI


domsj:
http://172.19.49.85/view/arakoon-1.7/job/arakoon-1.7-system-tests-slow-RIGHT/156/testReport/junit/arakoon_system_tests.server.right/system_tests_long_right/test_drop_master_with_load/
This test failed again and (I believe) it would be solved with the changes proposed here.

toolslive:
ok.

PR #280, opened 2013-12-17
Fix masterless cluster

domsj:
Looks good on CI.
These changes should not be forward ported to 1.7 (it already contains similar changes).
(They should be undone when forward porting 1.6 to 1.7)


toolslive:
sgoe.

PR #282, opened 2013-12-18
close of store and tlog shouldn't happen when an exception occured

domsj:
(backport from 7c6a7acb61d15830608528928c36df950249ba61)
Conflicts:
    src/node/node_main.ml


toolslive:
ok.

PR #285, opened 2013-12-20
slaves shouldn't ignore a master just because it's n is lower

domsj:
noticed the problem here: http://172.19.49.85/view/arakoon-1.7/job/arakoon-1.7-system-tests-slow-RIGHT/156/testReport/junit/arakoon_system_tests.server.right/system_tests_long_right/test_disable_tlog_compression/

Tests are still running
http://172.19.49.85/view/arakoon-git/job/arakoon-git-launcher/170/downstreambuildview/?
but not expecting any problems there


toolslive:
ok

PR #303, opened 2014-01-16
Flush store

domsj:
Should be good according to the tests Michael has run

### DIFF
--- a/extension/test/server/system_tests_common.py
+++ b/extension/test/server/system_tests_common.py
@@ -178,7 +178,7 @@ def flush_stores(nodes = None):
 
 def compare_stores( node1_id, node2_id ):
 
-    keys_to_skip = [ "*lease", "*i", "*master" ]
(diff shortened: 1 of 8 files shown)


src/client/python/Arakoon.py
```diff
@@ -235,7 +234,7 @@ def exists(self, key):
         """
         @type key : string
         @param key : key
-        @returen : True if there is a value for that key, False otherwise
+        @return : True if there is a value for that key, False otherwise
```
toolslive, line 237: ;)

extension/test/server/system_tests_common.py
```diff
@@ -162,9 +162,23 @@ def dump_store( node_id ):
 
     return dump_file
 
+def flush_store( node_id ):
+    client = get_client ()
+    client.allowDirtyReads()
+    client.setDirtyReadNode ( node_id )
+    client.range("", True, "a", False, 1)
```
toolslive, line 169: nifty.

toolslive:
ok.

PR #304, opened 2014-01-16
test drop master with a cluster that needs the previous master to make progress

domsj:
I think this is a better test, but maybe even better would be to test both situations?


toolslive:
fine.

PR #307, opened 2014-01-17
add Logger.ign_ macros

domsj:


NicolasT:
LGTM, but I'd like to hear from @toolslive whether this is a good idea / it's OK to have `ignore (log ...)` around.

PR #308, opened 2014-01-17
Slowdown collapse 1.7

domsj:


src/node/collapser.ml
```diff
@@ -108,7 +108,19 @@ let collapse_until (type s) (tlog_coll:Tlogcollection.tlog_collection)
                           Lwt.return ()
                       end
                       >>= fun () ->
-                      S.safe_insert_value new_store pi pv >>= fun _ ->
+                      begin
+                        match slowdown with
```
toolslive: it might be cleaner to do the match only once beforehand, and build a function to be used here.

toolslive:
looks ok.

PR #309, opened 2014-01-20
collapse threads should either successfully take the lock or fail immedi...

domsj:
...ately
(the thread should never hang while trying to acquire it)

I didn't actually see the problem happen, but while analysing a related failure for a little while I assumed this was the problem. It should be an improvement.


toolslive:
I agree (it should be an improvement)

PR #310, opened 2014-01-20
don't start any more client threads while going down

domsj:
I hope this can prevent the failure seen in http://172.19.49.85/view/arakoon-1.7/job/arakoon-1.7-system-tests-slow-SHAKY/173/testReport/junit/arakoon_system_tests.server.shaky/test_restart_master_long/test_restart_master_long/ .

### DIFF
--- a/src/tools/server.ml
+++ b/src/tools/server.ml
@@ -124,6 +124,13 @@ let make_server_thread
     let connection_counter = make_counter () in
     let client_threads = Hashtbl.create 10 in
     let _condition = Lwt_condition.create () in
+    let going_down = ref false in
+    let maybe_take () =
+      if !going_down
+      then
+        None
+      else
+        maybe_take () in
     let rec server_loop () =
       Lwt.catch
         (fun () ->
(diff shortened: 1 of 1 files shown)


src/tools/server.ml
```diff
@@ -188,11 +189,17 @@ let make_server_thread
...
-                   Lwt.catch
-                     (fun () -> t)
-                     (fun exn ->
-                        Logger.info_f_ ~exn "Exception in client thread %s" cid))
+                 if not !going_down
+                 then
+                   begin
+                     Hashtbl.add client_threads cid t;
+                     Lwt.catch
+                       (fun () -> t)
```
toolslive: let t = client_thread () in ..... (fun () -> t) 
is a bit heavy as construction.....

src/tools/server.ml
```diff
@@ -188,11 +189,17 @@ let make_server_thread
...
+                       (fun exn ->
+                        Logger.info_f_ ~exn "Exception in client thread %s" cid)
+                   end
+                 else
+                   Lwt.return ())
```
toolslive: maybe we want a variation of 'deny' here.

domsj:
I've tried to address the concerns with a new commit.
Tests are still running, but not expecting any problems there.

domsj:
Hmm, the commit message isn't really correct though .. should be `don't serve any more clients while going down, deny them instead` or sth along those lines.
Will rewrite history with a better message if we think the change should be like this...

toolslive:
ok.

PR #313, opened 2014-01-21
Implement in-place inject-as-head

NicolasT:


NicolasT:
```
test_inject_as_head : asserts shortcut for those who don't want to collapse periodically. (eta: 20 s) ... ok
arakoon_system_tests.server.shaky.test_inject_as_head.test_inject_as_head_current_corrupt ... ok
Test inject_as_head with a fake database file (should fail) ... ok
test_inject_as_head : asserts shortcut for those who don't want to collapse periodically. (eta: 20 s) ... ok
arakoon_system_tests.server.shaky.test_inject_as_head.test_inplace_inject_as_head ... ok
```

src/node/dump_store.ml
```diff
@@ -48,7 +48,7 @@ let dump_store filename =
 
 exception ExitWithCode of int;;
 
-let inject_as_head fn node_id cfg_fn force =
+let inject_as_head fn node_id cfg_fn force ~in_place =
```
domsj: Make force a labeled argument too?
The way you did it with ~in_place does add clarity at the call site.

src/node/dump_store.ml
```diff
@@ -108,8 +108,18 @@ let inject_as_head fn node_id cfg_fn force =
       | None -> failwith "can't happen"
       | Some i -> Sn.to_int (Tlc2.get_file_number i)
     in
-    Lwt_io.printf "cp %S %S" fn old_head_name >>=fun () ->
-    File_system.copy_file fn old_head_name true >>= fun () ->
+    begin
+      if (not in_place)
+      then begin
+        Lwt_io.printf "cp %S %S" fn old_head_name >>=fun () ->
+        File_system.copy_file fn old_head_name true
```
domsj: I know this line is the same as before, but maybe we should make the last argument here a labeled one too?

domsj:
Looks good

NicolasT:
Are you OK if I add the labels & the thing keeps compiling?

domsj:
Yes

NicolasT:
Superseded by #315.

PR #314, opened 2014-01-21
Drop master cluster maintenance

domsj:


src/paxos/master.ml
```diff
@@ -91,28 +91,42 @@ let stable_master (type s) constants ((n,new_i, lease_expire_waiters) as current
...
+                let log_e = ELog (fun () ->
+                                  "stable_master: half-lease_expired while doing drop-master, so not renewing lease")
+                in
+                Fsm.return ~sides:[log_e] (Stable_master current_state)
+            end
           else
             let log_e = ELog (fun () -> "stable_master: half-lease_expired: update lease." ) in
             let v = Value.create_master_value (me,0.0) in
-            let ms = {mo = None; v;n;i = new_i;lew = lease_expire_waiters} in
+            let ms = {mo = None; v;n;i = new_i;lew = []} in
```
NicolasT, line 118: Hmh, I'm not sure I understand what's going on here. Agree this should be `[]` when the `true` branch is taken above, but shouldn't the list be kept if the `false` branch is taken?

domsj, line 118: If I'm not mistaken then this branch is the one where `lease_expire_waiters == []`, while in the branch above (97-114) lease_expire_waiters is not empty. I was only trying to make that a bit more explicit by changing it to [].
In the branch above, when there are lease_expire_waiters (which requested drop master), first the LeaseExpired messages are ignored (110-113). Later on, if no other node takes over, lines 101-107 are executed. 101-107 could be considered the main change in this pull request.

NicolasT, line 118: Aaah, right, I see, wasn't immediately obvious.
Although @toolslive might disagree, I'd change 101/115 into `else (* is_empty lease_expire_waiters *)` to make this somewhat more clear (the 'if' block is fairly big, not used to such large contexts anymore).

domsj:
@NicolasT confirmed it's ok via skype

PR #320, opened 2014-01-24
fix calculation of lease expiration for small numbers (should start from...

domsj:
... a float)

### DIFF
--- a/src/paxos/multi_paxos.ml
+++ b/src/paxos/multi_paxos.ml
@@ -229,16 +229,17 @@ let start_lease_expiration_thread (type s) constants =
     let lease_start, slave = match S.who_master constants.store with
       | None -> 0.0, true
       | Some(m, ls) -> ls, constants.me <> m in
-    let sleep_sec =
+    let factor =
       if slave
       then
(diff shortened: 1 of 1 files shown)


src/paxos/multi_paxos.ml
```diff
@@ -229,16 +229,21 @@ let start_lease_expiration_thread (type s) constants =
...
-      then
-        (* sleep a little longer than necessary on a slave
-           to prevent a node thinking it's still the master
-           while some slaves have elected a new master amongst them
-           (in case the clocks don't all run at the same speed) *)
-        (float_of_int constants.lease_expiration) *. 1.1
-      else
-        float_of_int (constants.lease_expiration / 2) in
+      lease_expiration *.
+        begin
```
toolslive: (\* comment ... *)
let factor = if slave then ... else ... in
let sleep_sec = lease_expiration *. factor
might be easier on the brain.

domsj: yes, good point. I'll change it.

domsj:
updated

toolslive:
fine.

PR #322, opened 2014-01-24
Optimize tests

domsj:


extension/test/server/right/system_tests_long_right.py
```diff
@@ -469,15 +469,15 @@ def test_fsync():
...
-@Common.with_custom_setup(Common.setup_1_node, Common.basic_teardown)
+@Common.with_custom_setup(Common.setup_1_node_mini, Common.basic_teardown)
 def test_sabotage():
     """
-    scenario countering a sysadmin removing files (s)he shouldn't (eta : 160s)
+    scenario countering a sysadmin removing files (s)he shouldn't (eta : 16s)
     """
     clu = _getCluster()
-    tlog_size = Common.get_entries_per_tlog()
+    tlog_size = 1000
```
NicolasT, line 478: Would be nice to get this from the 'environment' setup by the decorator somehow.

extension/test/server/right/system_tests_long_right.py
```diff
@@ -469,15 +469,15 @@ def test_fsync():
...
     num_tlogs = 2
-    test_size = num_tlogs * tlog_size + 20
+    test_size = num_tlogs * tlog_size
```
NicolasT, line 480: Why drop the 20?

NicolasT:
LGTM overall

extension/test/server/right/system_tests_long_right.py
domsj, line 480: I thought these last 20 updates were unnecessary.
Actually I think `test_size = num_tlogs * tlog_size - number_of_paxos_values_due_to_election_and_nops + 1` should be enough here, unfortunately `number_of_paxos_values_due_to_election_and_nops` is a bit hard to determine ;-), but it should `>= 1`.

PR #324, opened 2014-01-31
timeout for drop_master (useful in a cluster doing maintenance)

domsj:


src/paxos/master.ml
```diff
@@ -59,52 +59,76 @@ let stable_master (type s) constants ((v',n,new_i, lease_expire_waiters) as curr
...
+              end
                 else
                   let log_e = ELog (fun () -> "stable_master: half-lease_expired: update lease." ) in
                   let v = Value.create_master_value (me,0L) in
                   let ff = fun _ -> Lwt.return () in
-		    (* TODO: we need election timeout as well here *)
+                  (* TODO: we need election timeout as well here *)
                   let ms = {mo = [ff];
                             v;n;i = new_i;
                             lew = lease_expire_waiters;
```
NicolasT: Shouldn't this be `[]` in line with the 1.7 change?

src/paxos/master.ml
```diff
@@ -59,52 +59,76 @@ let stable_master (type s) constants ((v',n,new_i, lease_expire_waiters) as curr
...
-		        begin
-                  let log_e = ELog (fun () -> Printf.sprintf "stable_master: handover to %s" p) in
-		          Fsm.return ~sides:[log_e] (Stable_master current_state)
-		        end
-	          else
-		        extend ()
-	        | _ -> extend()
-	    end
+          in
+          let extend () =
```
NicolasT: Consider renaming this to `maybe_extend` (I think that's what it's called in 1.7?). Now there are 2 `extend` functions defined, both with a different signature, where the second uses (but also aliases) the first, and is used afterwards, which is IMHO somewhat confusing.

domsj:
Addressed both comments

NicolasT:
LGTM

PR #327, opened 2014-02-04
Pyrakoon

toolslive:


pylabs/Compat.py
```diff
@@ -14,6 +14,26 @@ class Status:
...
 
+VAR = 'ARAKOON_PYTHON_CLIENT'
+def _normal_client():
+    logging.info("opting for normal client")
+    global arakoon_client
+    from arakoon import Arakoon
+    arakoon_client = Arakoon
+
+if os.environ.has_key(VAR):
+    global arakoon_client
```
NicolasT, line 25: No need for this `global` declaration.

pylabs/Compat.py
```diff
@@ -14,6 +14,26 @@ class Status:
...
+    wanted = os.environ[VAR]
+    if wanted == 'pyrakoon':
+        logging.info("opting for pyrakoon")
+        print "pyrakoon"
+        from pyrakoon import compat
+        arakoon_client = compat
+    else:
+        _normal_client()
```
NicolasT, line 33: Why put this case in a function and the other case toplevel?

pylabs/test/server/quick/system_tests_basic.py
```diff
@@ -109,7 +106,13 @@ def test_max_value_size_tinkering ():
     time.sleep(1.0)
     C.assert_running_nodes(1)
     client = C.get_client()
-    assert_raises (ArakoonException, client.set, key, value)
+    try:
+        client.set(key, value)
+        assert_true(False)
+    except Exception, e:
+        # with multiple clients, we have multiple ArakoonException hierarchies
+        assert_equals(e.__class__.__name__, 'ArakoonException')
```
NicolasT, line 114: :worried:

pylabs/test/server/quick/system_tests_basic.py
```diff
@@ -197,15 +200,14 @@ def test_aSSert_scenario_1():
...
+    assert_raises( X.arakoon_client.ArakoonAssertionFailed, client.aSSert, 'x', None)
 
 @C.with_custom_setup(C.default_setup, C.basic_teardown)
 def test_aSSert_scenario_3():
     client = C.get_client()
     client.set('x','x')
-    ass = arakoon.ArakoonProtocol.Assert('x','x')
-    seq = arakoon.ArakoonProtocol.Sequence()
-    seq.addUpdate(ass)
+    seq = client.makeSequence()
```
NicolasT, line 209: I guess this shows some API difference between the `arakoon` package and `pyrakoon.compat`, where the latter is supposed to be API-compatible with the former. As such, shouldn't `pyrakoon.compat` be updated somehow?

pylabs/test/server/quick/system_tests_basic.py
```diff
@@ -340,6 +341,7 @@ def test_drop_master_singleton():
         raise Error
     except Exception, e:
         rc = e.args[0]
+        ARA_ERR_NOT_SUPPORTED = 32
```
NicolasT, line 344: Same as above, I guess this points at some API difference.

Having these magic numbers embedded in some test code might not be the right approach.

pylabs/test/server/shaky/test_sso_deployment.py
```diff
@@ -34,6 +34,7 @@
 
 @Common.with_custom_setup( Common.setup_1_node_forced_master, Common.basic_teardown )
 def test_sso_deployment():
+    pass
```
NicolasT, line 37: Leftover?

src/client/python/ArakoonProtocol.py
```diff
@@ -145,6 +145,13 @@ def getNodeLocations(self, nodeId):
         """
         return self._nodes[ nodeId ]
 
+    def getNodeLocation(self, nodeId):
```
NicolasT, line 148: I'm not sure this is the best place to put this...
- It's not entirely clear what this method does compared to `getNodeLocations` (consider adding a docstring at least)
- Is this method useful from a user perspective?
- It changes API, so the `pyrakoon.compat` API should be updated accordingly.

pylabs/test/server/quick/system_tests_basic.py
toolslive, line 209: it is. (see pull request over there)

pylabs/test/server/quick/system_tests_basic.py
toolslive, line 344: alternative is to expose these constants in both (with a standard name)

pylabs/test/server/shaky/test_sso_deployment.py
toolslive, line 37: nope. deliberate. nose will now list the test name iso the doc string.

src/client/python/ArakoonProtocol.py
toolslive, line 148: This makes both ClientConfig classes interchangeable for both clients. I'm not sure this is the best approach neither.

pylabs/Compat.py
toolslive, line 33: I use the function twice. it's just about not repeating code.
(I could reorder the tests as we want _normal_client 3 cases out of 4)

pylabs/test/server/quick/system_tests_basic.py
toolslive, line 114: will fix.

PR #331, opened 2014-02-11
add stats for range_entries

toolslive:


toolslive:
CI ok.

domsj:
Looks good.
Add the same (statistics/client/cli) for rev_range(_entries) now that we're thinking of it?

domsj:
Ok looks good. A bit strange that travis failed though (I don't think your changes caused the failure).

PR #332, opened 2014-02-12
master_dictate: send accept messages to slaves before the master accepts...

domsj:
... (and fsyncs) the paxos value

(this removes the fsync on the master from the critical path)

Ok on CI


toolslive:
owkee.

PR #333, opened 2014-02-12
Batched store optimize range queries

domsj:
Looks good on CI

### DIFF
--- a/pylabs/extensions/arakoon_ext/server/ArakoonManagement.py
+++ b/pylabs/extensions/arakoon_ext/server/ArakoonManagement.py
@@ -81,11 +81,11 @@ class ArakoonManagement:
 class ArakoonCluster:
 
     def __init__(self, clusterName):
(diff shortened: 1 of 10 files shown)


domsj:
Hmm, only thinking of it now, but this does break https://github.com/Incubaid/arakoon/blob/1.7/pylabs/test/server/system_tests_common.py?source=c#L179 . (Of which strangely enough we see no noticeable effects when running our test suite, but amplidata does.)
So we might want to make flushing of the store an operation which can be explicitly triggered by clients?

toolslive:
maybe; but only exposed in mgmt api?

domsj:
Ok, exposed the flushStore call in the management API.
CI succeeded.
Also opened a PR for pyrakoon: https://github.com/Incubaid/pyrakoon/pull/5 .

toolslive:
ok

PR #336, opened 2014-02-12
META and debian changelog

toolslive:


META
```diff
@@ -1,7 +1,7 @@
 # -*- conf -*-
 
 description = "Arakoon client"
-version = "1.6.8"
+version = "1.6.9"
```
NicolasT, line 4: This got me wondering: do we need new `arakoon-client` releases when only _node_ things changed?

debian/changelog
```diff
@@ -1,3 +1,7 @@
+arakoon (1.6.9) unstable; urgency=low
+  * new upstream version
+
+ -- Romain Slootmaekers  <romain@incubaid.com>  Wed, 12 Feb 2014 16:00:00 +0100
```
NicolasT: I guess the date will need to be adapted?

toolslive: do we care?

domsj: I like it better when it's correct, and it's only a small effort. (You can even do it straight from github.)
So in short: yes, I care.

PR #337, opened 2014-02-12
immediately trigger lease_expired after catchup if no new master lease w...

domsj:
...as learned

take n from master when receiving an Accept from it for an already learned value

### DIFF
--- a/src/paxos/slave.ml
+++ b/src/paxos/slave.ml
@@ -190,7 +190,7 @@ let slave_steady_state (type s) constants state event =
                 Logger.debug_f_ "%s: slave, received old accepted (latest applied to store), ignoring" constants.me >>= fun () ->
                 let reply = Accepted(n',i') in
                 let send_e = ESend(reply, source) in
-                Fsm.return ~sides:[ send_e ] (Slave_steady_state state)
+                Fsm.return ~sides:[ send_e ] (Slave_steady_state (n', i, maybe_previous))
               end
             else
               accept_value None n' i' v "steady_state :: replying again to previous with %S"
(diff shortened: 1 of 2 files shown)


domsj:
CI looks good. Drop master tests succeeded twice.
Also they now only take about 60% of the time they did before, so looks like an improvement to me.

src/paxos/slave.ml
```diff
@@ -190,7 +190,7 @@ let slave_steady_state (type s) constants state event =
                 Logger.debug_f_ "%s: slave, received old accepted (latest applied to store), ignoring" constants.me >>= fun () ->
                 let reply = Accepted(n',i') in
                 let send_e = ESend(reply, source) in
-                Fsm.return ~sides:[ send_e ] (Slave_steady_state state)
+                Fsm.return ~sides:[ send_e ] (Slave_steady_state (n', i, maybe_previous))
```
NicolasT, line 193: Care to explain? As in: does this fix a bug?

domsj, line 193: I guess it depends on your definition of a bug :-).

This code path is taken when a node receives an Accept message for a value it already learned (which is already applied to the store), from a new master (with a higher n).
This situation can happen when a drop master is ongoing. Both slaves send prepares. One receives a promise from the other, and starts sending accepts (the accept being handled here). The other receives a Nak from the master and does a catchup. After the catchup it has already learned the value for which it receives this accept.
If this node now immediately takes the n coming in the accept message it will start following the new master.
If it doesn't take the n (as it was before this branch), then the next accept message from the master will be ignored in https://github.com/Incubaid/arakoon/blob/fix-drop-master/src/paxos/slave.ml#L197 , resulting in a longer time before a new master can really establish itself.

Another option would be to leave this code as is and add a match case `| Accept(n', i', v) when i' = i && n' > n -> ...` so the message isn't just ignored there.

So this change doesn't fix any 'wrong' behavior, but it does make the transition to a new master smoother.

domsj:
@toolslive said merge it in on skype

PR #338, opened 2014-02-13
Fuse client protocol

domsj:


extension/test/server/system_tests_common.py
```diff
@@ -1164,6 +1164,22 @@ def range_entries_scenario( start_suffix ):
         logging.info("on failure moment, master was: %s", client._masterId)
         raise ex
 
+def heavy_range_entries_scenario( start_suffix, count, queries ):
+    value_format_str_ = "p" * 4200 +  "value_%012d"
+    client = get_client()
+
+    for i in range(count):
```
NicolasT: `xrange`

extension/test/server/system_tests_common.py
```diff
@@ -1164,6 +1164,22 @@ def range_entries_scenario( start_suffix ):
...
+        suffix = ( i + start_suffix )
+        key = key_format_str % suffix
+        value = value_format_str_ % suffix
+
+        client.set(key, value)
+
+
+    start_key = key_format_str % (start_suffix )
+    end_suffix = key_format_str % (start_suffix + count)
+    for i in range(queries) :
```
NicolasT: `xrange`

src/client/client_protocol.ml
```diff
@@ -145,8 +145,13 @@ let handle_sequence ~sync ic oc backend =
 
   end
 
-let one_command (ic,oc,id) (backend:Backend.backend) =
-  read_command (ic,oc) >>= function
+let one_command stop (ic,oc,id) (backend:Backend.backend) =
+  read_command (ic,oc) >>= fun command ->
+  if !stop
+  then
+    Lwt.return true
```
NicolasT, line 152: If I'm not mistaken this will cause the client connection to be closed, but I guess a client will try to reconnect immediately, which would result in a reconnection loop (as long as the server thread is running).

Maybe 'sleep' for some time first (and hope the server process ends before the nap is over)?

domsj, line 152: Hmm, didn't really think of this.
Sleeping is an option, another would be to do something like in 1.7: https://github.com/Incubaid/arakoon/blob/1.7/src/tools/server.ml?source=c#L174 ?
So when the flag here is true, no new connections can be accepted either...

domsj:
Addressed comments but waiting for CI again now.

NicolasT:
I think there could be better ways to handle the connections (not just dropping them), but LGTM for now.

PR #339, opened 2014-02-14
slave: append to tlog + reply with accepted before applying the consensu...

domsj:
...s effect (so that it is not part of the critical loop)

waiting for http://172.19.49.85/view/arakoon-git/job/arakoon-git-launcher/215/downstreambuildview/? (should probably be ok)

### DIFF
--- a/src/paxos/master.ml
+++ b/src/paxos/master.ml
@@ -34,7 +34,10 @@ let master_consensus constants {mo;v;n;i;lew} () =
   let con_e = EConsensus(mo, v,n,i) in
   let log_e = ELog (fun () ->
     Printf.sprintf "on_consensus for : %s => %i finished_fs (in master_consensus)"
-      (Value.value2s v) (List.length mo) )
+      (Value.value2s v)
+      (match mo with
+       | None -> 0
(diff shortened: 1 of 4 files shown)


src/paxos/master.ml
```diff
@@ -34,7 +34,10 @@ let master_consensus constants {mo;v;n;i;lew} () =
   let con_e = EConsensus(mo, v,n,i) in
   let log_e = ELog (fun () ->
     Printf.sprintf "on_consensus for : %s => %i finished_fs (in master_consensus)"
-      (Value.value2s v) (List.length mo) )
+      (Value.value2s v)
+      (match mo with
+       | None -> 0
+       | Some mo -> List.length mo))
```
NicolasT, line 40: Mss stdlib uitbreiden met `maybe : 'b -> ('a -> 'b) -> 'a option -> 'b` en dan `maybe 0 List.length mo` :wink:

NicolasT:
I think it'd be good for @toolslive to take a look as well, but LGTM.

toolslive:
ok.

PR #340, opened 2014-02-18
bs_bench compiles again (+ random gets)

toolslive:
it's a bit 'ad hoc' but I think we want to take it along nonetheless


domsj:
Ok

PR #341, opened 2014-02-18
1.6 for 1.7

domsj:
This should contain all the latest 1.6 changes.
There were many merge conflicts, so please look closely.
CI tests are still running, but should be ok. I'll post an update when they finish.


toolslive:
ok.

PR #343, opened 2014-02-21
Wait for client protocol thread

domsj:


src/tools/server.ml
```diff
@@ -39,11 +39,15 @@ let close = function
...
 
-let deny (ic,oc,cid) =
+let deny_max (ic,oc,cid) =
   Logger.warning_ "max connections reached, denying this one" >>= fun () ->
   Llio.output_int oc 0xfe >>= fun () ->
   Llio.output_string oc "too many clients"
 
+let deny_closing (ic,oc,cid) =
+  Logger.warning_ "closing socket, denying this one" >>= fun () ->
+  Llio.output_int oc 0xfe >>= fun () ->
```
NicolasT: Wouldn't `E_GOING_DOWN` be more suitable?

NicolasT:
Maybe it'd be useful to add a rationale to the commit message of 7422f1b? It's fairly obvious what the change does, but not why it's necessary (i.e. what's wrong with the earlier approach, and why the messaging service needs no waiting anymore).

domsj:
Actually I'm (partially) changing my mind on this ... I think we shouldn't wait on either of those threads.
We start closing the store, and some clients receive E_GOING_DOWN from somewhere in server.ml.
If we don't wait in node_main.ml and immediately go to close the store, then some other clients (already connected and in the process of doing a request) might receive the E_GOING_DOWN from inside of store.ml. No problems there.
Closing the store while operations are still operating on it shouldn't be a problem either.
I'll add another commit.

toolslive:
LGTM

PR #349, opened 2014-02-26
improve check to determine if lease expiration can be fired immediately

domsj:
Still waiting for http://172.19.49.85/view/arakoon-git/job/arakoon-git-launcher/223/downstreambuildview/?

I think it should prevent failures such as the ones seen here: http://172.19.49.85/view/arakoon-git/job/arakoon-amplidata-git-system-tests-slow-RIGHT/57/testReport/


toolslive:
fine.

PR #350, opened 2014-02-27
also check the lease was previously expired if maybe injecting LeaseExpi...

domsj:
...red after catchup

I think this is what I really really want ;-).


toolslive:
doe maar.

PR #351, opened 2014-03-03
Memstore range cursor

domsj:
This pull request changes:
- interface of simple_store -> exposes cursors
- range like queries are implemented in store
- mem_store exposes cursors (with the help of Map.Cursor & Map.Zipper)
- removed test_backend, this exposed 2 minor bugs (ping in sync_backend, set_interval in store), and 1 bug in the tests
- optimized handling of range-like queries (no List.rev or Array.to_list stuff, note that some of these are hidden inside camltc)

Edit: re-running CI again after my final changes, see http://172.19.49.85/view/arakoon-git/job/arakoon-git-launcher/230/


domsj:
Forgot to mention this in the original description:
Delete_prefix is still in the simple_store interface. Handling it differently is an explicit non-goal of this PR.
Mostly because I'm not yet sure about what is the best way to implement this. (I know tokyo cabinet has cur_out, but the situation is different for mem_store and leveldb.)
Slightly related is the fact that currently the cursor queries take in account changes from a currently executing transaction.. TC doesn't like changes to the database while walking over a cursor though, so still need to be careful. I'm a proponent of changing those semantics though.

src/lib/std.ml
```diff
@@ -0,0 +1,391 @@
...
+module List = struct
+  include List
+
+  let is_empty = function
+    | [] -> true
+    | _ -> false
+
+  let find' test list = try Some(find test list) with Not_found -> None
+end
+
```
domsj, line 228: Changes in this file start here. (The rest I copied...)

src/client/remote_nodestream_test.ml
```diff
@@ -1,8 +1,29 @@
+(*
+This file is part of Arakoon, a distributed key-value store. Copyright
+(C) 2010 Incubaid BVBA
```
toolslive, line 3: shouldn't the template be adapted?

src/lib/std.ml
```diff
@@ -0,0 +1,391 @@
+(*
+This file is part of Arakoon, a distributed key-value store. Copyright
+(C) 2013 Incubaid BVBA
```
toolslive, line 3: hehe.

src/lib/std.ml
```diff
@@ -0,0 +1,391 @@
...
+
+      let first ?(z=[]) t =
+        let rec inner acc = function
+          | Empty ->
+             None
+          | Node(Empty, _, _, _, _) as t ->
+             Some (t, acc)
+          | Node(l, _, _, _, _) as t ->
+             inner ((t, Left)::acc) l in
+        inner z (Obj.magic t)
```
toolslive, line 258: hm.... don't know exactly what to think of this.

src/lib/std.ml
```diff
@@ -0,0 +1,391 @@
...
+        | None ->
+           failwith "invalid cursor"
+        | Some z ->
+           Zipper.get z
+
+    end
+  end
+end
+
+type 'a counted_list = (int * 'a list)
```
toolslive, line 391: would have preferred a record. 
you will be constructing/deconstructing a zillion tuples.

src/node/local_store.ml
```diff
@@ -261,98 +234,57 @@ let relocate ls new_location =
...
+
+let with_cursor ls f =
+  let bdb = Camltc.Hotc.get_bdb ls.db in
+  B.with_cursor bdb (fun _ cur -> f (bdb, cur))
+
+let wrap_not_found f =
+  try
+    f ();
+    true
+  with Not_found -> false
```
toolslive, line 251: like said in the 'chat', we want to modify camltc semantics to bool iso exeptions too.

src/tools/llio.ml
```diff
@@ -232,11 +232,14 @@ let input_string_list ic = input_list input_string ic
...
+  output_int oc count >>= fun () ->
+  Client_log.debug_f "Outputting list with %d elements" count >>= fun () ->
+  Lwt_list.iter_s (output_element oc) list
+
 let output_list output_element oc list =
   let n = List.length list in
-  output_int oc n  >>= fun () ->
-  Client_log.debug_f "Outputting list with %d elements" n >>= fun () ->
-  Lwt_list.iter_s (output_element oc) list
+  output_counted_list output_element oc (n, list)
```
toolslive: this type of construction is wrong (imo)
If you really want to have code sharing here, you should build a function that writes out a list when you now the length.
then use it with let n = (List.length list) in output_x  n list 
and let (n,list) = cl in output_x n list

src/client/remote_nodestream_test.ml
domsj, line 3: Hah, guess it should.
I would prefer to do this for all files in another branch/PR though.

src/lib/std.ml
domsj, line 258: I know Obj.magic isn't really desired...
Other options would be
- propose these (or similar) changes to be picked up in the std lib
- copy the map implementation here

I prefer the first of those.

src/lib/std.ml
domsj, line 391: Is constructing tuples more expensive than constructing records?
I would think (assume) they have the same cost...

src/tools/llio.ml
domsj: That's not an insensible opinion ;)
I'll update.

src/node/local_store.ml
domsj, line 251: Uhu. I'll make an update to camltc soon.

PR #352, opened 2014-03-04
Lcnum tests

domsj:


cfg/arakoon.ini
```diff
@@ -9,6 +9,8 @@ cluster = arakoon_0 ,arakoon_1
 # and 'preferred_master' is implicit
 #preferred_masters = arakoon_0, arakoon_1
 #lease_period = 60 
+lcnum=1024 
+ncnum=1024
```
domsj, line 13: Do we want to put the same values here as the defaults defined in node_cfg.ml ?

domsj:
I'm sorry, I should not have opened this as a pull request so impulsively.

PR #354, opened 2014-03-07
Consistency

toolslive:
extending true/false for allowDirty with a third alternative: at_least(i)


src/client/python/ArakoonProtocol.py
```diff
@@ -488,6 +489,35 @@ def _recvStringOption ( con ):
...
+    pass
+
+class NoGuarantee(Consistency):
+    def __init__(self):
+        self._v = _packBool(True)
+    
+    def isDirty(self):
+        return True
+    
+class AtLeast(Consistency):
```
domsj, line 517: return True for isDirty

src/client/python/ArakoonProtocol.py
```diff
@@ -488,6 +489,35 @@ def _recvStringOption ( con ):
...
+        
+    def isDirty(self):
+        return False
+
+    def __str__(self):
+        return self.__class__.__name__
+
+    __repr__ = __str__
+
+class Consistent(Consistency):
```
domsj, line 507: also initialize self._v and implement isDirty here?

domsj:
I would like the name get_mark more for the client method, and correspondingly a file mark.ml (iso the current stamp.ml)

domsj:
I think adding a python test which does a read with At_least(a_high_mark_which_will_surely_be_inconsistent) would be a good idea.
It verifies that 1) the python encoding of this is correct and 2) the server correctly responds with E_INCONSISTENT_READ.

domsj:
Hmm, I'm wondering if the mark should also be aware of the j value in the database?

PR #357, opened 2014-03-07
update all copyright notices

domsj:
and sneakily remove an obsolete file


domsj:
updated

toolslive:
fine. so we'll have to do this again next year?

domsj:
Dunno, you're the one who complained about the wrong year number in another PR.
For all I care we can also remove the year number...

toolslive:
we can also remove the headers and replace it with one single file stating the licensing under src/

domsj:
Hmm .. that would indeed be better.
Now I'm wondering how other projects handle this, will have to do some research.
Maybe iso removing the headers replace them with a pointer to that file?

PR #360, opened 2014-03-10
seed with opam 1.1 lwt-2.4.4 etc

toolslive:


domsj:
there are 2 other seed files too: `/jenkins/base/seed` and `/jenkins/kwik/seed`

PR #366, opened 2014-03-17
fix intermittently failing base catchup tests

domsj:
This should fix http://172.19.49.85/view/arakoon-1.7/job/arakoon-1.7-base/231/ .
I admit it's probably not the best way, any other suggestions are welcome.


toolslive:
ok.

PR #367, opened 2014-03-17
more aggressive election timeout when in stable master state

domsj:
have a look at the logs related to http://172.19.49.85/view/arakoon-1.7/job/arakoon-1.7-system-tests-slow-RIGHT/244/testReport/junit/arakoon_system_tests.server.right/system_tests_long_right/test_drop_master/


toolslive:
ok.

PR #368, opened 2014-03-17
less creation of tuples.

toolslive:
let x, pos2 = Llio.x_from s pos in 
gets reworked to
let x = Llio.x_from buf in


toolslive:
travis has issues as I use primitives only available in 4.01

domsj:
ok, then update travis file to no longer build with 4.00.1

domsj:
looks good (and quite mechanic)

PR #373, opened 2014-03-19
test suites based on test_it.py

toolslive:


jenkins/B/070_run_tests.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash -xue
+
+echo $PWD
+
+python test_it.py --with-xunit --xunit-file=${PWD}/B.xml \
+  server.left.system_tests_long_left \
+  server.left.test_collapse \
+  server.left.test_concurrent_collapse_fails \
+  server.left.test_copy_db_to_head
```
domsj, line 9: Can we specify these as server.left.\* ? (And similar for the others)

PR #377, opened 2014-03-31
remove unused name attribute in node config

domsj:


toolslive:
ok, didn't this used to be a one-liner?

domsj:
Yea, sth like that. with ?w=1 it now says `Showing 3 changed files with 4 additions and 11 deletions`.
The extra changes are because 1) I had to fix some tests and 2) there was an additional place the name attribute got set.

PR #379, opened 2014-04-02
User hooks2

domsj:


src/client/common.ml
```diff
@@ -430,18 +437,18 @@ let delete_prefix (ic,oc) prefix =
   request oc outgoing >>= fun () ->
   response ic Llio.input_int
 
+let rec change_to_update = function
```
toolslive: let open Arakoon_client in ?

src/node/batched_store.ml
```diff
@@ -186,13 +179,9 @@ struct
     _verify_tx s tx;
     _track_or_apply_vo s k (Some v)
 
-  let delete s tx k =
+  let put s tx k vo =
```
toolslive, line 182: all for it.

src/plugins/registry.mli
```diff
@@ -14,19 +14,44 @@ See the License for the specific language governing permissions and
...
 *)
 
+class type cursor_db =
+  object
+    method get_key : unit -> Key.t
+    method get_value : unit -> string
+    method jump : string -> bool
+    method last : unit -> bool
+    method next : unit -> bool
+    method prev : unit -> bool
```
toolslive, line 24: this is a good bare-bones api, but do we want a comfort api  on top of this too?

toolslive:
ok.

src/client/common.ml
domsj: Yes, will update

src/plugins/registry.mli
domsj, line 24: Yea, agreed that we want to provide some comfort too.
I'm thinking of a helper module with `jump'` `fold_range` and `fold_rev_range` a la the helper module in Simple_store
Currently the one in simple_store works against Simple_store.t types and it allows iterating all keys, not just the public keys like this one does... so it requires some (small) changes.
Anyway, I'll have a look tomorrow.

domsj:
added the comfort + a test

toolslive:
nice.

PR #380, opened 2014-04-02
make server_loop a REAL tail call again

toolslive:
got fooled by compiler annotations claiming it to be a tail call no more.


domsj:
fine

PR #389, opened 2014-04-15
really produce highest prio event

domsj:


src/paxos/multi_paxos_fsm.ml
```diff
@@ -655,7 +645,15 @@ let rec paxos_produce buffers
        Logger.debug_f_ "%s: T:waiting for event (%s)" me wmsg >>= fun () ->
        let t0 = Unix.gettimeofday () in
 
-       Lwt.pick waiters >>= fun ready_list ->
+       let wait_for_buffer = function
+         | Client -> Lwt_buffer.wait_for_item buffers.client_buffer
+         | Node -> Lwt_buffer.wait_for_item buffers.node_buffer
+         | Inject -> Lwt_buffer.wait_for_item buffers.inject_buffer
+         | Election_timeout -> Lwt_buffer.wait_for_item buffers.election_timeout_buffer in
```
toolslive, line 652: let lookup_buffer = function 
 | Client -> buffers.client_buffer
 | ...
in 
...
?

toolslive:
ok.

src/paxos/multi_paxos_fsm.ml
domsj, line 652: I tried, but that forces all `'a Lwt_buffer.t`'s to have the same `'a` ... not sure how I can avoid that? Any hints would be appreciated...
Heh, a little magic could probably help but I would rather avoid that :)

PR #390, opened 2014-04-16
use List.assoc iso manual match

toolslive:


src/client/arakoon_exc.ml
```diff
@@ -34,40 +34,30 @@ type rc =
...
-  | 0x10l -> E_GOING_DOWN
-  | 0x20l -> E_NOT_SUPPORTED
-  | 0x21l -> E_NO_LONGER_MASTER
-  | 0x80l -> E_INCONSISTENT_READ
-  | 0xfel -> E_MAX_CONNECTIONS
-  | 0xffl -> E_UNKNOWN_FAILURE
-  | _     -> E_UNKNOWN_FAILURE
+let int32_of_rc rc = List.assoc rc mapping
+
+let rev_mapping = List.map (fun (a,b) -> (b,a) ) mapping
```
domsj: If we're going towards this way of defining things (which is fine by me), then I would prefer for the lookup (mapping and rev_mapping) to use Hashtbl's iso association lists.

toolslive: I considered this (also considered a map) but my reasoning is that we're in an exceptional path so it doesn't matter anyway.

domsj: E_OK, E_NOT_FOUND and E_ASSERTION_FAILED ain't exceptional to me (ok maybe some people would discuss about those last 2... but I don't share their opinion)

domsj:
sgoe

PR #391, opened 2014-04-16
1.7 for 1.8

domsj:


META
```diff
@@ -1,7 +1,7 @@
 # -*- conf -*-
 
 description = "Arakoon client"
-version = "1.7.1"
+version = "1.7.2"
```
toolslive, line 4: oops

domsj, line 4: yea, that still has to be updated on the 1.8 branch

toolslive:
fine. I'll do the 1.8 in META in a separate pr.

PR #394, opened 2014-04-23
Check for ENOSPC conditions at startup

NicolasT:


NicolasT:
Maybe this should go against 1.7 instead?

NicolasT:
Running CI.

domsj:
looks good

PR #396, opened 2014-04-24
1.7 for 1.8

NicolasT:
CI: `arakoon-git-launcher` 259

### DIFF
--- a/src/client/client_protocol.ml
+++ b/src/client/client_protocol.ml
@@ -88,40 +88,33 @@ let handle_exception oc exn=
   begin
     match exn with
       | XException(Arakoon_exc.E_NOT_FOUND, msg) ->
-        Lwt.return (Arakoon_exc.E_NOT_FOUND,msg, false, false, Logger.Debug)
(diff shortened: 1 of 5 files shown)


domsj:
This looks really familiar... ok when CI is ok

NicolasT:
@domsj There's a failure in slow-right 3075: test_drop_master ("Could not determine the master node"). CI hickup?

domsj:
Yea, I've seen that one quite a few times lately; I think I fixed it in #397 .
If you look at http://10.101.186.2/jenkins/jobs/jenkins-arakoon-git-system-tests-slow-RIGHT-3075/home/jenkinslxc/qbase3/var/tmp/arakoon_system_tests/test_drop_master/sturdy_0/log/ you can see that sturdy_0 became the master, which happens to be the node being stopped by https://github.com/Incubaid/arakoon/blob/1.7/pylabs/test/server/right/system_tests_long_right.py#L335 , that finally results in the error you saw.
So if that's all then pls go ahead and merge this in...

PR #397, opened 2014-04-25
prevent drop master test from failing with NoMaster because in the

domsj:
beginning the active master was stopped


NicolasT:
I'd write this slightly differently (determine which node to stop first, then issue a single 'stop' call), but LGTM.

PR #399, opened 2014-04-29
Master lease

domsj:
I think this works as intended (see http://jira.incubaid.com/browse/ARAKOON-468 ) .. but I want some comments about how to organize it.

I'm not really satisfied with the RenewLease module in node_main which allows creating a few `unit Lwt.t`s which should then be ran.
I was thinking maybe these loops should be written as 'actors' ?
(I could add a few helpers which allow easily defining a few actors)
Instead of returning the unit Lwt.t these could be `Lwt.ignore_result`ed, maybe that's cleaner?

I could move the RenewLease module to another file (probably somewhere in the /paxos dir).

Feedback wanted.

### DIFF
--- a/pylabs/test/server/quick/system_tests_basic.py
+++ b/pylabs/test/server/quick/system_tests_basic.py
@@ -104,7 +104,7 @@ def test_max_value_size_tinkering ():
     C.assert_running_nodes(1)
     client = C.get_client()
     assert_raises (X.arakoon_client.ArakoonException, client.set, key, value)
-    
+
 
 @C.with_custom_setup(C.setup_1_node,C.basic_teardown)
 def test_marker_presence_required ():
@@ -385,7 +385,7 @@ def test_consistency():
     m = client.get_txid()
     logging.debug("m = %s", m)
     assert_equals(str(m).find("AtLeast"),0)
-    time.sleep(5)
+    client.nop()
     client.setConsistency(m)
     v = client.get('x')
     assert_equals(v,'X')
--- a/src/msg/tcp_messaging.ml
+++ b/src/msg/tcp_messaging.ml
@@ -359,7 +359,7 @@ class tcp_messaging
         let (source:id) = Llio.string_from buffer in
         let target      = Llio.string_from buffer in
         let msg         = Message.from_buffer buffer in
-        Logger.debug_f_ "message from %s for %s" source target >>= fun () ->
+        (* Logger.debug_f_ "message from %s for %s" source target >>= fun () -> *)
         if drop_it msg source target then Lwt.return b1
         else
           begin
--- a/src/msg/tcp_messaging.ml
+++ b/src/msg/tcp_messaging.ml
@@ -83,7 +83,7 @@ class tcp_messaging
 
     method register_receivers mapping = List.iter (fun (id,address) -> self # _register_receiver id address) mapping
 
-    method private _target' sub_target target =
+    method private _target' sub_target ~target =
       match sub_target with
         | None -> target
         | Some sub -> Printf.sprintf "%s\n%s" target sub
@@ -127,8 +127,7 @@ class tcp_messaging
       let tq = self # _get_send_q ~target in
       Lwt_buffer.add (source, target, sub_target, m) tq
 
-    method private get_buffer ?sub_target (target:id) =
-      let target' = self # _target' sub_target target in
+    method private _get_buffer target' =
       try Hashtbl.find _qs target'
       with | Not_found ->
         begin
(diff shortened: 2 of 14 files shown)


src/msg/tcp_messaging.ml
```diff
@@ -83,9 +83,18 @@ class tcp_messaging
 
     method register_receivers mapping = List.iter (fun (id,address) -> self # _register_receiver id address) mapping
 
+    method private _target' sub_target ~target =
+      match sub_target with
+        | None -> target
+        | Some sub -> Printf.sprintf "%s\n%s" target sub
```
NicolasT, line 89: I didn't look further down yet, so maybe my comment doesn't make any sense, but: it feels like `(string, string option)` or `type target = Target of string | SubTarget of (string * string)` could be a better approach? Unless it really needs to be a string somehow.

src/node/node_main.ml
```diff
@@ -267,6 +267,183 @@ let build_service_ssl_context me cluster =
...
+    | Timeout of (float)
+    | Kick
+    | DropMaster
+
+  module NodeSet = Set.Make(String)
+
+  open Message
+
+  let make_payload counter =
+    let buf = Buffer.create 20 in
```
NicolasT: An int64 is always 8 bytes, no need to allocate a 20-byte buffer.

src/node/node_main.ml
```diff
@@ -267,6 +267,183 @@ let build_service_ssl_context me cluster =
...
+  let decode_granted_msg from msg =
+    let payload = Message.payload_of msg in
+    Grant (from, decode_payload payload)
+
+  let start_master_renew_lease_loop
+        (type s) (module S : Store.STORE with type t = s) store
+        (messaging : Messaging.messaging)
+        me others lease_expiration stop =
+
+    (* let me = constants.Multi_paxos.me in *)
```
NicolasT: Remove commented code?

src/node/node_main.ml
```diff
@@ -267,6 +267,183 @@ let build_service_ssl_context me cluster =
...
+    (* let others = constants.Multi_paxos.others in *)
+    (* let lease_expiration = float constants.Multi_paxos.lease_expiration in *)
+    let new_granteds = NodeSet.add me NodeSet.empty in
+    let quorum = ((List.length others) / 2) + 1 in
```
NicolasT: If I'm not mistaken a quorum can be configured as well. Shouldn't this value be taken into account?

src/paxos/multi_paxos.ml
```diff
@@ -63,7 +63,7 @@ let network_of_messaging (m:messaging) =
   let send msg source target =
     Logger.debug_f_ "%s: sending msg to %s: %s" source target (Mp_msg.MPMessage.string_of msg) >>= fun () ->
     let g = MPMessage.generic_of msg in
-    m # send_message g ~source ~target
+    m # send_message g ~source ?sub_target:None ~target
```
NicolasT, line 66: No need to pass 'None' to an optional argument?

src/msg/tcp_messaging.ml
domsj, line 89: Currently target is serialized as a string. So far I did not want to break that, that's why I'm stuffing the subtarget in there too.
I think we can break backwards compatibility in the internode protocol... that would make nicer solutions possible

src/node/node_main.ml
domsj: Good point. While I was working on this I had some other info in here too so the 20 made more sense.
Will update.

src/node/node_main.ml
domsj: Ok

src/node/node_main.ml
domsj: Yes, that should probably be taken into account. I'll pass in the needed info...

src/paxos/multi_paxos.ml
domsj, line 66: Yea, that's what one would think, but then it gives a type error:

```
File "src/paxos/multi_paxos.ml", line 66, characters 4-38:
Error: This expression has type ?sub_target:string -> unit Lwt.t
       but an expression was expected of type 'a Lwt.t
Command exited with code 2.
```

Passing in 'None' was the easiest solution

pylabs/test/server/quick/system_tests_basic.py
```diff
@@ -385,7 +385,7 @@ def test_consistency():
     m = client.get_txid()
     logging.debug("m = %s", m)
     assert_equals(str(m).find("AtLeast"),0)
-    time.sleep(5)
+    client.nop()
```
domsj, line 388: Ideally this change should not be needed...
Maybe make learning a value on the slaves explicit with `Learn of n * i` messages? These messages could then be sent from master to slave when the master has reached consensus on -- aka learned -- a value.

src/paxos/multi_paxos.ml
NicolasT, line 66: Hmh, interesting. Must be related to mixture & order of labeled & optional arguments?

PR #400, opened 2014-05-02
Disallow running as root unless really stubborn

NicolasT:


domsj:
Sorry for leaving this hanging for such a long time ... I'm still a bit unsure about this one.
I'm totally pro not running arakoon as root, but the same could be said about not running with fsync=true (which is an even worse thing to do imo).
So I think that we want to handle both cases in the same way ... wether that's the warning in the log or refusing to start -- unless a certain env variable or tainted config setting is present.
Btw is there a reason here to prefer an env variable over a tainted config setting?

NicolasT:
I wouldn't implement this using some warning in a log: nobody will ever see that anyway.
I used an environment variable instead of a configuration setting because
- Unlike the `fsync` thing, this isn't related to how an Arakoon node operates at runtime. It's meant to prevent any runtime at all.
- I figure some would without any hesitation add `__tainted_run_as_root = true` to their Arakoon configuration file template. Managing environment variables might be slightly harder depending on the deployment system in use.
- When using the work-around, it's more clear at the location of the exec call one is doing something fishy.

domsj:
By the same token nobody is going to see the fsync warning in the log

NicolasT:
I won't disagree there :wink:

domsj:
So do you think we should be less lenient in the fsync case?
I know it's not exactly the same thing but I'm interested in your opinion.

PR #401, opened 2014-05-02
Regularly dump TCP_INFO statistics of outgoing connections

NicolasT:
This introduces TCP-level connection statistics logging of outgoing inter-node connections/sockets. Using this information, connectivity issues (excessive packet loss, huge round-trip times,...) can be detected.

```
May  2 17:14:27 0613: (tcp_messaging|info): TCP info for 127.0.0.1:4011: { state=1; ca_state=0;
retransmits=0; probes=0; backoff=0; options=7; snd_wscale=7; rcv_wscale=7; rto=201000; ato=0;
snd_mss=21888; rcv_mss=536; unacked=0; sacked=0; lost=0; retrans=0; fackets=0; last_data_sent=35;
last_ack_sent=0; last_data_recv=335849803; last_ack_recv=35; pmtu=65535; rcv_ssthresh=43690;
rtt=1000; rttvar=500; snd_ssthresh=2147483647; snd_cwnd=10; advmss=65483; reordering=3; rcv_rtt=0;
rcv_space=43690; total_retrans=0; }
```

### DIFF
--- a/src/libcutil.clib
+++ b/src/libcutil.clib
@@ -3,3 +3,4 @@ bsd_crc32c.o
 cpudetect.o
 arakoon_limits.o
 cllio_c.o
+tcp_info_stubs.o
--- a/src/msg/tcp_messaging.ml
+++ b/src/msg/tcp_messaging.ml
@@ -23,7 +23,7 @@ open Log_extra
 open Lwt_buffer
 open Network
 
-type connection = Lwt_io.input_channel * Lwt_io.output_channel
+type connection = Lwt_unix.file_descr * Lwt_io.input_channel * Lwt_io.output_channel
 
 let section =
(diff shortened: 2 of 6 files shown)


domsj:
Ok, the way I've understood this is that it logs information which can't be captured from outside arakoon, so I'm in favor of this addition.
A few remarks:
- do we want this on 1.7?
- add this to the regular statistics instead?

NicolasT:
For 1.7: not sure, I don't think it's that important (nobody actually asked for this :wink:)
Adding to regular statistics: I'd vote against. Rationale:
- This would result in the data to be passed in a client `statistics` call result as well, which seems rather useless: it's hard to correlate these connection stats to Arakoon client call stats
- The information is not very useful except in very specific situations for (node) system administrators

PR #406, opened 2014-05-09
Remove dead code

NicolasT:
Some removals might be controversial...

### DIFF
--- a/src/node/catchup.ml
+++ b/src/node/catchup.ml
@@ -84,21 +84,6 @@ let _with_client_connection ~tls_ctx (ips,port) f =
   loop ips
 
 
-let compare_store_tlc store tlc =
-  tlc # get_last_i () >>= fun tlc_i ->
-  let m_store_i = store # consensus_i () in
-  match m_store_i with
-    | None ->
-      if tlc_i = ( Sn.succ Sn.start )
-      then Lwt.return Store_1_behind
-      else Lwt.return Store_n_behind
-    | Some store_i when store_i = tlc_i -> Lwt.return Store_tlc_equal
-    | Some store_i when store_i > tlc_i -> Lwt.return Store_ahead
-    | Some store_i ->
-      if store_i = (Sn.pred tlc_i)
-      then Lwt.return Store_1_behind
-      else Lwt.return Store_n_behind
-
 let head_saved_epilogue hfn tlog_coll =
   (* maybe some tlogs must be thrown away because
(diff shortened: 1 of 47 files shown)


NicolasT:
Comments welcome, though I won't merge this yet, first enable & fix some more warnings-as-errors.

NicolasT:
These changes no longer only 'remove dead code':
- Remove some dead code (according to coverage testing)
- Enable more warnings, most 'as error', and fix corresponding errors
- Re-instate coverage builds

src/node/statistics.ml
```diff
@@ -241,18 +241,6 @@ module Statistics = struct
...
-    let x = new_op t start in
-    update_x_stats t.range_time_stats x
-
-  let new_range_entries t (start:float) =
-    let x = new_op t start in
-    update_x_stats t.range_time_stats x
-
-  let new_rev_range_entries t (start:float) =
-    let x = new_op t start in
-    update_x_stats t.rev_range_entries_time_stats x
```
domsj, line 254: Hmm, don't these indicate that those statistics should still be captured? (Probably in sync_backend if I'm not mistaken)

src/node/catchup.ml
```diff
@@ -84,21 +82,6 @@ let _with_client_connection ~tls_ctx (ips,port) f =
...
-    | None ->
-      if tlc_i = ( Sn.succ Sn.start )
-      then Lwt.return Store_1_behind
-      else Lwt.return Store_n_behind
-    | Some store_i when store_i = tlc_i -> Lwt.return Store_tlc_equal
-    | Some store_i when store_i > tlc_i -> Lwt.return Store_ahead
-    | Some store_i ->
-      if store_i = (Sn.pred tlc_i)
-      then Lwt.return Store_1_behind
-      else Lwt.return Store_n_behind
```
domsj, line 100: The type `store_tlc_cmp` can now be removed too

domsj:
Some minor comments but I definitely like cleaning up (y)

NicolasT:
Reworking against 1.7.

src/node/statistics.ml
NicolasT, line 254: They are: there are other functions with the same names (but different types) defined below, so these ones weren't used at all.

src/node/catchup.ml
NicolasT, line 100: Will look into this in the forthcoming PR.

PR #410, opened 2014-05-14
remove client lib oUnit and quickcheck dependency

domsj:


NicolasT:
OK if the test script worked.

PR #411, opened 2014-05-14
fix __tainted_fsync_tlog_dir typo

domsj:
Tests are still running for this one, but the change is so trivial it's almost impossible I made a mistake :)


NicolasT:
LGTM

PR #413, opened 2014-05-14
Cleanup myocamlbuild

NicolasT:


myocamlbuild.ml
```diff
@@ -1,14 +1,22 @@
+(* Findlib dependencies *)
+let dependencies = [ "lwt"
+                   ; "oUnit"
+                   ; "camltc"
+                   ; "snappy"
+                   ; "ssl"
+                   ; "bz2"
```
domsj, line 7: add quickcheck?

NicolasT, line 7: Good point, I thought that was only used in 3.x. Will add.

NicolasT, line 7: Done.

domsj, line 7: hmm, I think bz2 should be camlbz2.
now --version lists the following on my system:

```
bz2     0.6.0   [n/a]
```

however ... filling in camlbz2 doesn't work because Findlib.query can't find it??

NicolasT, line 7: That's because the package is called `camlbz2` in Opam, but the actual OCaml/`ocamlfind` package name is `bz2`, hence e.g. in `_tags`: 

```
$ grep bz2 _tags
true: package(bz2)
```

Also:

```
$ ls ~/.opam/4.01.0/lib/camlbz2
$ ls ~/.opam/4.01.0/lib/bz2
bz2.a  bz2.cma  bz2.cmi  bz2.cmx  bz2.cmxa  bz2.mli  libcamlbz2.a  META
```

domsj:
LGTM

PR #414, opened 2014-05-14
Remove `Cllio` and friends

NicolasT:


domsj:
Ok, cleaning up is always nice

PR #415, opened 2014-05-14
Remove `Cpu_detect` and friends

NicolasT:
The only required functionality (detection of SSE4.2 support) is merged into
`arakoon_crc32c.c`


domsj:
Ok, I suppose this is good

PR #417, opened 2014-05-15
Update `arakoon.odocl`

NicolasT:
The modules listed in `arakoon.odocl` were no longer up-to-date. Some were
listed but are no longer present, others were missing.

This commit updates the file, so API documentation can be built again using

```
$ ocamlbuild -use-ocamlfind arakoon.docdir/index.html
```

The file was generated using

```
$ find src/ -iname '*.ml' -exec basename {} .ml \; | \
    sed 's/./\U&/' | \
    grep -v -e _test$ \
            -e Bs_bench \
            -e OUnit_XML \
            -e Plugin_update_max \
            -e Profile \
            -e Test_utils | \
    sort > arakoon.odocl
```


domsj:
Aha, some people might like this.
LGTM

PR #420, opened 2014-05-15
Remove unused `String_extra` module

NicolasT:


toolslive:
is used in nursery, and 3.0 (prefix_match)

NicolasT:
Indeed it is in 3.0 (so a forward merge will need some care). Not sure about nursery?

```
[nicolas@chi arakoon (1.7 %)]$ git show-ref refs/heads/1.7
891800e81dae4046f2e4286e6b0e6ef176ad60d9 refs/heads/1.7
[nicolas@chi arakoon (1.7 %)]$ git grep prefix_match
src/tools/string_extra.ml:let prefix_match prefix k =
```

domsj:
Another option is moving that function to Std.String

NicolasT:
Agree (in 3.0).

domsj:
Ok, LGTM

PR #421, opened 2014-05-16
Turn `str` arguments into `[str]` as soon as possible

NicolasT:


domsj:
Thanks, LGTM

domsj:
I don't yet know on which branch they want the fix...

PR #422, opened 2014-05-16
fix qshell extension for node ips list

domsj:
Still waiting on CI, but feel free to review already


NicolasT:
LGTM if CI is OK & manual validation works.

PR #423, opened 2014-05-16
Custom optimize

domsj:


src/node/local_store.ml
```diff
@@ -286,4 +265,80 @@ let make_store ~lcnum ~ncnum read_only db_name  =
...
+          Bdb.first source cursor;
+          let max = Some batch_size in
+          let rec loop i =
+            let count =
+              Bdb.copy_from_cursor
+                ~source
+                ~cursor
+                ~target
+                ~max in
+            if !stop
```
toolslive, line 298: if !stop then false => not stop && ...
(and maybe replace stop with proceed:  proceed & ....)

src/node/local_store.ml
```diff
@@ -286,4 +265,80 @@ let make_store ~lcnum ~ncnum read_only db_name  =
...
+               lcnum;}
+
+
+let optimize ls ~quiesced ~stop =
+  let open Camltc in
+  let batch_size = 10_000 in
+  let log_size = 1_000_000 in
+  let log_batch_count = (log_size / batch_size) - 1 in
+  let db_optimal = ls.location ^ ".opt" in
+  File_system.exists db_optimal >>= fun opt_exists ->
```
NicolasT: This (and the `unlink` later on) is (mostly in theory, since nobody is supposed to create or remove any files in Arakoon dirs except for the running process) a [TOCTOU](http://en.wikipedia.org/wiki/Time_of_check_to_time_of_use) bug. There's a `safe_unlink` utility in `Node_cfg` which works around this. Consider moving that to `File_system` and use it?

src/node/local_store.ml
```diff
@@ -286,4 +265,80 @@ let make_store ~lcnum ~ncnum read_only db_name  =
...
+              end
+            else
+              true
+          in
+          loop 0)
+     in
+     Lwt_preemptive.detach go () >>= fun finished ->
+        if finished
+        then
+          Camltc.Hotc.sync ~detached:true ls_optimal.db >>= fun () ->
```
NicolasT: Is there a reason to run this in another detached thread, instead of handling this at the end of `go`?

src/node/sync_backend.ml
```diff
@@ -633,7 +633,7 @@ struct
       method optimize_db () =
         Logger.info_ "optimize_db: enter" >>= fun () ->
         let mode = Quiesce.Mode.ReadOnly in
-        self # try_quiesced ~mode (fun () -> S.optimize store) >>= fun () ->
+        self # try_quiesced ~mode (fun () -> Lwt.map ignore (S.optimize store)) >>= fun _ ->
```
NicolasT: I might be mis-interpreting this, but isn't it funny to `map` `ignore` on some action result so it yields `()`, then pattern-match on `_` instead of `()` as before?
Basically, map `ignore` _or_ match on `_`.

NicolasT:
Looks fine.

src/node/local_store.ml
domsj: Waiting on https://github.com/toolslive/camltc/pull/20 to make the suggested change possible.

src/node/local_store.ml
domsj, line 298: although `bool && ...` is clever and compact I'm not really convinced it helps readability.

we're using such a stop flag in a lot of places in arakoon .. if we want to change this we might want to do that everywhere (and in another PR)?

domsj:
Fwiw, I added another commit here, which CI agreed with. May I suppose this is still looking fine?

src/node/local_store.ml
toolslive, line 298: if you're not convinced then it doesn't (help readability).

NicolasT:
OK

PR #424, opened 2014-05-19
fix test_restart_master_long

domsj:


NicolasT:
LGTM, but I wonder why we didn't run into this before?

PR #426, opened 2014-05-19
removed a few hundred unused vars.

toolslive:


src/client/arakoon_remote_client_test.ml
```diff
@@ -29,7 +29,7 @@ let _CLUSTER = "sweety"
 
 type real_test = Arakoon_client.client -> unit Lwt.t
 
-let __client_server_wrapper__ cluster (real_test:real_test) =
+let __client_server_wrapper__ _cluster (real_test:real_test) =
```
NicolasT: Consider removing the argument instead?

src/node/catchup.ml
```diff
@@ -105,7 +105,7 @@ let stop_fuse stop =
   else
     Lwt.return ()
 
-let catchup_tlog (type s) ~tls_ctx ~stop me other_configs ~cluster_id  mr_name ((module S : Store.STORE with type t = s),store,tlog_coll)
+let catchup_tlog (type s) ~tls_ctx ~stop _me other_configs ~cluster_id  mr_name ((module S : Store.STORE with type t = s),store,tlog_coll)
```
NicolasT: Consider removing the argument instead, if the interface permits?

src/node/node_main.ml
```diff
@@ -651,7 +651,7 @@ let _main_2 (type s)
       let unlock_killswitch (_:int) = Lwt_mutex.unlock killswitch in
       let listen_for_signal () = Lwt_mutex.lock killswitch in
 
-      let start_backend stop (master, constants, buffers, new_i, store) =
+      let start_backend stop (master, constants, buffers, new_i, _store) =
```
NicolasT: Consider pruning the tuple?

src/node/sync_backend.ml
```diff
@@ -243,7 +242,7 @@ struct
         self # _check_interval [key] >>= fun () ->
         let () = assert_value_size value in
         let update = Update.Set(key,value) in
-        let update_sets (update_result:Store.update_result) = Statistics.new_set _stats key value start in
+        let update_sets (_update_result:Store.update_result) = Statistics.new_set _stats key value start in
```
NicolasT, line 245: Consider turning into `()`?

src/node/sync_backend.ml
```diff
@@ -291,11 +290,13 @@ struct
         _update_rendezvous self update no_stats push_update ~so_post
 
       method aSSert ~consistency (key:string) (vo:string option) =
+        ignore consistency;
```
NicolasT: I'm a bit surprised this can be ignored.

src/node/sync_backend.ml
```diff
@@ -323,7 +324,7 @@ struct
           | Some w -> assert_value_size w
         in
         let update = Update.Replace(key,wanted) in
-        let update_stats ur = Statistics.new_replace _stats start in
+        let update_stats __ur = Statistics.new_replace _stats start in
```
NicolasT: Typo?

src/node/sync_backend.ml
```diff
@@ -350,10 +352,10 @@ struct
       method delete key = log_o self "delete %S" key >>= fun () ->
         let start = Unix.gettimeofday () in
         let update = Update.Delete key in
-        let update_stats ur = Statistics.new_delete _stats start in
+        let update_stats __ur = Statistics.new_delete _stats start in
```
NicolasT: Typo?

src/nursery/routing.ml
```diff
@@ -117,7 +117,7 @@ module Routing = struct
 *)
 
   let remove t c_id =
-    let rec inner_remove t c_id other_side =
+    let rec inner_remove t c_id _other_side =
```
NicolasT: Consider removing the argument?

src/tools/crash_logger.ml
```diff
@@ -31,7 +31,7 @@ type e = {
 let circ_buf = Lwt_sequence.create()
 let msg_cnt = ref 0
 
-let add_to_crash_log section level msgs =
+let add_to_crash_log _section level msgs =
```
NicolasT: Consider removing the argument?

src/node/sync_backend.ml
toolslive: it's an update .... but the api needs to remain as is.

src/node/sync_backend.ml
toolslive: does it matter?

src/tools/crash_logger.ml
toolslive: For a lot of these, there is a _feeling_ of an api. ie the log api has a section a level and msgs.... I would like to keep the signature, even though we don't use the parameters.

src/node/sync_backend.ml
domsj: Why does the API need to remain as is?

src/tools/crash_logger.ml
domsj: It might make sense to dump the section to the crash log too?

src/node/node_cfg.ml
```diff
@@ -405,7 +404,7 @@ module Node_cfg = struct
     let tlx_dir =
       let rec _find = function
         | [] -> tlog_dir
-        | x :: xs -> try get_string "tlf_dir"
+        | _ :: xs -> try get_string "tlf_dir"
```
domsj: This non-usage actually indicates a bug

src/node/node_main.ml
```diff
@@ -303,7 +303,7 @@ module X = struct
       Lwt.return r
     end
 
-  let on_accept (type s) statistics (tlog_coll:Tlogcollection.tlog_collection) (module S : Store.STORE with type t = s) store (v,n,i) =
+  let on_accept (type s) statistics (tlog_coll:Tlogcollection.tlog_collection) (module S : Store.STORE with type t = s) _store (v,n,i) =
```
domsj: store and (module S) arg can be removed instead

src/node/sync_backend.ml
toolslive: ok, seems to be consensus the parameter can be removed from the signature, and it will remain in the protocol only.

domsj: On second thought I think the api should remain as is, but we should no longer consider the operation to be an update but treat it as a query instead (and depending on ~consistency maybe allow the query to happen on a slave too).
Same for assert_exists below

toolslive: ok. will leave it as is; turning this into a query can be done in a different PR.

domsj:
LGTM

PR #427, opened 2014-05-19
Fix multiple masters

domsj:


NicolasT:
LGTM

NicolasT:
@domsj could you handle forward-merge?

domsj:
Yes, will do

PR #429, opened 2014-05-23
prevent 2 nodes in wait_for_accepteds state remaining isolated (by ignor...

domsj:
...ing each other)

You may want to add &w=1 when comparing the change


NicolasT:
LGTM, and :+1: for improving the 'obvious exhaustiveness' of the cases given the guard clauses.

PR #430, opened 2014-05-23
Fix for catchup-only.

domsj:
Previously when the get_db call succeeded the result of this was not
copied to head.db. This results in incomplete data on this specific
node. I removed this optimization so the node only does regular
catchup.


toolslive:
fine.

PR #432, opened 2014-05-27
Cleanup warnings

domsj:


domsj:
Still waiting for CI

domsj:
Fyi: There's a failing test related to a not found user_function (which is definitely an error in the test case), but now I'm left wondering how we want to handle that.
(The error in the test case being exposed now is due to the change in exception handling I introduced in store.ml.)

domsj:
Ok, so I changed the behaviour of user_functions as discussed on the chat.
Things look ok on CI, travis failure is camltc related.

toolslive:
it's a dirty job but I'm glad somebody did it. LGTM

PR #434, opened 2014-05-27
Non lwtey

domsj:
This PR
- removes unneeded lwt-ness
- turn aSSert and aSSert_exists into queries
- fix a concurrency bug in confirm

### DIFF
--- a/src/client/client_protocol.ml
+++ b/src/client/client_protocol.ml
@@ -145,46 +145,48 @@ let one_command stop (ic,oc,id) (backend:Backend.backend) =
   then
     Lwt.return true
   else
+    let wrap_exception f =
+      Lwt.catch
+        (fun () ->
+         try f ()
+         with exn -> Lwt.fail exn)
+        (handle_exception oc) in
     match command with
   | PING ->
      begin
-       Lwt.catch
+       wrap_exception
          (fun () ->
           Llio.input_string ic >>= fun client_id ->
           Llio.input_string ic >>= fun cluster_id ->
           Logger.debug_f_ "connection=%s PING: client_id=%S cluster_id=%S" id client_id cluster_id >>= fun () ->
           backend # hello client_id cluster_id >>= fun msg ->
           response_ok_string oc msg)
-         (handle_exception oc)
     end
   | FLUSH_STORE ->
      begin
-       Lwt.catch
+       wrap_exception
          (fun () ->
           Logger.debug_f_ "connection=%s FLUSH_STORE" id >>= fun () ->
           backend # flush_store () >>= fun () ->
           response_ok_unit oc)
-         (handle_exception oc)
      end
   | EXISTS ->
     begin
       Common.input_consistency ic  >>= fun consistency ->
       Llio.input_string ic >>= fun key ->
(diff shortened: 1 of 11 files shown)


src/client/client_protocol.ml
```diff
@@ -145,46 +145,48 @@ let one_command stop (ic,oc,id) (backend:Backend.backend) =
   then
     Lwt.return true
   else
+    let wrap_exception f =
+      Lwt.catch
+        (fun () ->
+         try f ()
```
NicolasT: I think this try-with is done by Lwt already?

src/client/client_protocol.ml
```diff
@@ -492,123 +475,112 @@ let one_command stop (ic,oc,id) (backend:Backend.backend) =
...
           response_ok_unit oc )
-        (handle_exception oc)
     end
   | GET_KEY_COUNT ->
     begin
-      Lwt.catch
+      wrap_exception
         (fun() ->
            Logger.debug_f_ "connection=%s GET_KEY_COUNT" id >>= fun () ->
            backend # get_key_count () >>= fun kc ->
```
NicolasT: I'm somewhat surprised `get` etc are non-Lwt, but `get_key_count` is.

src/client/client_protocol.ml
```diff
@@ -492,123 +475,112 @@ let one_command stop (ic,oc,id) (backend:Backend.backend) =
...
     end
   | CONFIRM ->
     begin
       Llio.input_string ic >>= fun key ->
       Llio.input_string ic >>= fun value ->
-      Lwt.catch
+      wrap_exception
         (fun () ->
            Logger.debug_f_ "connection=%s CONFIRM: key=%S" id key >>= fun () ->
            backend # confirm key value >>= fun () ->
```
NicolasT, line 574: I'm somewhat surprised `get` etc are non-Lwt, but `confirm` is.

src/client/client_protocol.ml
domsj: Hmm, apparently you're right

src/node/dump_store.ml
```diff
@@ -21,10 +21,10 @@ open Interval
 
 module S = (val (Store.make_store_module (module Batched_store.Local_store)))
 
-let try_fetch name (f:unit -> 'a Lwt.t) (r2s: 'a -> string)  =
+let try_fetch name (f:unit -> 'a) (r2s: 'a -> string)  =
   Lwt.catch
     (fun () ->
-       f () >>= fun r ->
+       Lwt.wrap f >>= fun r ->
```
NicolasT: Is this `wrap` instead of simply calling?

domsj: I could write `let r = f () in` instead. I wrongly assumed the wrap was needed here, but Lwt.catch has an implicit wrap already...

src/client/client_protocol.ml
domsj, line 574: confirm might do an update, so this should remain Lwt.t

src/client/client_protocol.ml
domsj: hmm, somehow 'missed' this one

domsj:
Updated, but I'll run CI again to be sure.

domsj:
CI agrees with my latest changes here. Travis failure seems to be camltc related.

PR #435, opened 2014-05-30
1.7 for 1.8

domsj:


NicolasT:
Nothing jumps out, LGTM if CI is OK.

PR #436, opened 2014-05-30
cleanup more unused variables

domsj:


src/paxos/master.ml
```diff
@@ -257,6 +257,8 @@ let stable_master (type s) constants ((n,new_i, lease_expire_waiters) as current
 
 let master_dictate constants ms () =
   let {mo;v;n;i;lew} = ms in
+  ignore mo;
```
NicolasT: Can't you not match on them instead?

domsj: that gives another warning/error instead...

domsj: I could write `let v,n,i = ms.v, ms.n, ms.i in` if you prefer that

NicolasT: I thought `let { v; n; i; _ } = ms in` is supposed to work and not generate warnings/errors?

domsj: You're right, updated.

NicolasT:
LGTM

PR #437, opened 2014-05-30
robustify test_start_stop_wrapper

domsj:
Sometimes this test fails ...


NicolasT:
Hmm... Not very fond of this type of bandainds. Wouldn't it be better to try connecting to the node in a loop (just a simple socket connection), which might even take more than 1 second, but (hopefully) less?

domsj:
Yea that would indeed be better. This is just a quick/lazy fix.
That behaviour should then be used in a bunch of other places in the tests too (search for `time.sleep` and you'll see where, I spot about 7 of these in system_tests_basic.py alone)

PR #438, opened 2014-06-02
allow scenarios for benchmarks

toolslive:
a simple way to specify what you want as part of the benchmark scenario

### DIFF
--- a/src/client/benchmark.ml
+++ b/src/client/benchmark.ml
@@ -14,8 +14,6 @@ See the License for the specific language governing permissions and
 limitations under the License.
 *)
 
-
-
 open Lwt
 open Arakoon_client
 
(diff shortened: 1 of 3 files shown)


src/client/benchmark.ml
```diff
@@ -297,11 +299,13 @@ let benchmark
...
-  Lwt.join (ts phase_6) >>= fun () ->
-  Lwt.return ()
+  let phase_map = [
+    ("master" , phase_0); ("set", phase_1);
+    ("set_tx", phase_2); ("get", phase_3); ("multi_get", phase_4);
+    ("range", phase_5);("range_entries", phase_6)
+  ]
+  in
+
+  let phase_of_name n = List.assoc n phase_map in
```
NicolasT: Maybe catch `Not_found` and re-raise a more descriptive failure?

src/main/main.ml
```diff
@@ -195,6 +195,7 @@ let main () =
   and linc = ref true
   and right = ref None
   and rinc = ref false
+  and scenario = ref "master, set, set_tx, get, multi_get, range, range_entries"
```
NicolasT: Maybe push `phase_map` in `benchmark.ml` to the top level, and construct this `scenario` string based on the keys found in there, so this default is always in sync with 'all benchmarks', a la `String.concat ", " (List.map fst Benchmark.phase_map)`

NicolasT:
LGTM

PR #439, opened 2014-06-04
Use a temporary file when receiving `head.db`

NicolasT:
During catchup, `head.db` was overwritten in-place, which is rather
unsafe at several levels.

This commit changes the receiving mechanism to use a temporary file, and
safely rename the received file (if successful) to the target `head.db`.

### DIFF
--- a/src/tlog/tlc2.ml
+++ b/src/tlog/tlc2.ml
@@ -658,12 +658,23 @@ class tlc2
       Logger.info_ "save_head()" >>= fun () ->
       Llio.input_int64 ic >>= fun length ->
       let hf_name = self # get_head_name () in
-      Lwt_io.with_file
-        ~flags:[Unix.O_WRONLY;Unix.O_CREAT]
-        ~mode:Lwt_io.output
-        hf_name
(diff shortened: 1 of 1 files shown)


NicolasT:
Didn't pass through CI yet, launching now.

NicolasT:
`arakoon-git-launcher` 315

NicolasT:
CI passed, all blue.

NicolasT:
The question whether this approach is required was raised, due to the increased storage requirements in the `head_dir`. A different approach was proposed, which boiled down to basically removing `head.db` as well as all tlogs on a system whenever the transfer of a new `head.db` from another node is required during catchup (which implies all tlogs are too old anyway).

Whilst I understand this might be a concern, I am convinced an Arakoon node should at no time remove any information it contains, even though this information is very outdated and available on (a quorum of) other systems in the cluster. As such I think the approach implemented by this patch is the correct one.

mark-christiaens:
I vote for Nicolas solution to go in.  It's safer than the existing code and the most defensive solution.  If anyone is convinced that we should go further then we can do another pull request.

domsj:
Ok lgtm, merge it in

PR #441, opened 2014-06-05
Configure `__tainted_tlog_entries_per_file` also for `-catchup-only`

NicolasT:


NicolasT:
Not sure CI is required for this change. The manual failure scenario discovered yesterday now passes.

domsj:
LGTM

PR #444, opened 2014-06-12
Optimize `Llio`

NicolasT:
See https://github.com/Incubaid/arakoon/commit/224dbc9fee05bfd4600d7c965dca61168a423a82.patch for a rationale.

Will launch CI first.

### DIFF
--- a/src/tools/llio.ml
+++ b/src/tools/llio.ml
@@ -39,11 +39,6 @@ let lwt_failfmt fmt =
   let k x = Lwt.fail (Failure x) in
   Printf.ksprintf k fmt
 
-let (<: ) = Int32.shift_left
-let (<::) = Int64.shift_left
-let (>: ) = Int32.shift_right_logical
-let (>::) = Int64.shift_right_logical
-
 type buffer = {buf : string ; mutable pos : int }
 let make_buffer buf pos = {buf;pos}
 let buffer_done b = b.pos = String.length b.buf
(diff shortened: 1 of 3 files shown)


src/tools/llio.ml
```diff
@@ -65,17 +62,9 @@ let int32_from b =
...
-    let mask = Int32.of_int 0xff <: pos in
-    let code = (Int32.logand i32 mask) >: pos in
-    Char.chr (Int32.to_int code)
-  in
-  let add i = Buffer.add_char buffer (char_at i) in
-  add 0;
-  add 1;
-  add 2;
-  add 3
+  let s = String.create 4 in
```
domsj, line 65: I think we can avoid allocating this string every time this function gets called.
There could be 1 target string used by both int32_to and int64_to. (Not sure if that target string could be shared by input/output_int32/64, as those are lwt-ey...)
The gain from this change is probably marginal though.

NicolasT, line 65: @mark-christiaens made the same comment yesterday. Assuming non-preempted 'single-threaded' execution this might be true, but I'm reluctant to rely on such things... If ever OCaml would support a parallel runtime, this could introduce hard-to-pinpoint bugs.

mark-christiaens, line 65: Another strategy might be to add a target string as parameter.  It's then up to the caller to re-allocate or reuse a target string, leaving it up to the caller to run the risk of races.

toolslive, line 65: There already  something that's _passed in_ : the (buffer : Buffer.t). So if you really want to fiddle it's better to replace the type of the buffer. Then again, it might be better to review the protocol to reduce the number of string copies altogether. Without a typical use case it's hard to see what exactly the value is of such an optimization effort.

NicolasT:
CI (`arakoon-git-launcher` 317) is all blue.

src/tools/llio.ml
domsj, line 65: The serialization could indeed be done in entirely different ways, e.g. by not using a Buffer.t but instead sth along the lines of

```
type t =
| String of string
| Int64 of Int64.t
| Cons of t * t
| ...
```

and then write that to an output channel using io vectors...

But I propose not making perfect the enemy of good here.
I think that in case you want to go multi-threaded there's a lot more to be checked anyway, so I would still favor omitting the allocation, but if you want to keep it like this then that's fine for me too.
Maybe we should start annotating where we should be careful / make changes when/if we go multi-threaded?
Anyway, with or without the extra string allocation: the biggest part of this optimization is already realized... so LGTM

PR #445, opened 2014-06-12
don't keep the lease_start in the store, always initialize it to the

domsj:
current time such that a node will respect another master for at least
an entire lease_period when it wakes up.

### DIFF
--- a/src/node/simple_store.ml
+++ b/src/node/simple_store.ml
@@ -25,8 +25,8 @@ let __j_key = "*j"
 let __interval_key = "*interval"
 let __routing_key = "*routing"
 let __master_key  = "*master"
-let __lease_key = "*lease"
-let __lease_key2 = "*lease2"
+(* let __lease_key = "*lease" *)
+(* let __lease_key2 = "*lease2" *)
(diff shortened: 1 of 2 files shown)


NicolasT:
Will this also behave the same when there is no `*master` in the store?

domsj:
I think so, that's the

```
     with Not_found ->
       None
```

path which I left unchanged

src/node/simple_store.ml
```diff
@@ -25,8 +25,8 @@ let __j_key = "*j"
 let __interval_key = "*interval"
 let __routing_key = "*routing"
 let __master_key  = "*master"
-let __lease_key = "*lease"
-let __lease_key2 = "*lease2"
+(* let __lease_key = "*lease" *)
+(* let __lease_key2 = "*lease2" *)
```
toolslive, line 29: why the comments?

domsj, line 29: I thought it might prevent future contributors from reusing that key (maybe for the same purpose which would not be a good idea). Otoh, who am I kidding ;)?

PR #446, opened 2014-06-13
Remove custom GC configuration

NicolasT:
During some non-real-world benchmarking (single-node cluster, `fsync`
disabled, running in `tmpfs`) it turned out the custom GC configuration
as set up by Arakoon has a non-negligible performance impact: the `set`
scenario of the built-in benchmark goes from +- 115s to +- 93s.

It looks like the default GC settings (of OCaml 4.01.0) are better
suited for the Arakoon workload.

Next to this, the custom settings removed the ability to set
site-specific configurations using `OCAMLRUNPARAM`.

### DIFF
--- a/src/node/node_main.ml
+++ b/src/node/node_main.ml
@@ -19,7 +19,6 @@ open Tcp_messaging
 open Update
 open Lwt
 open Lwt_buffer
-open Gc
 open Master_type
 open Client_cfg
 open Statistics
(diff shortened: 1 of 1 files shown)


NicolasT:
Will run CI and report how this influences the runtime.

toolslive:
**historical note**: These settings were engineered (trial, error and bisect) to make the _monkey_ stay under 128MB while it rattled its cage. At that time, this was considered more important than raw performance. LGTM

NicolasT:
Thanks for that clarification :+1: Luckily we're not running 128MB nodes anymore, and if one does, `OCAMLRUNPARAM` can come to the rescue :smiley:

PR #447, opened 2014-06-13
Fix typo in `Benchmark`

NicolasT:
It's surprising to see `get ... took ...` messages printed when one is
only executing the `set` benchmark.


domsj:
LGTM

PR #449, opened 2014-06-16
update bs_bench for latest camltc changes + add rule to Makefile

domsj:


Makefile
```diff
@@ -14,6 +14,9 @@ clean:
 build:
 	ocamlbuild -j 4 -use-ocamlfind arakoon.byte arakoon.native arakoon_client.cma arakoon_client.cmxa arakoon_client.a arakoon_client.cmxs
 
+bench:
+	ocamlbuild -use-ocamlfind bs_bench.native
```
NicolasT, line 18: Since this rule doesn't produce its target, this should be `.PHONY` :wink:

domsj, line 18: Fixed

NicolasT:
LGTM

PR #450, opened 2014-06-17
sync store before closing tlog

domsj:


src/node/store.ml
```diff
@@ -290,10 +289,10 @@ struct
   let flush store =
     S.flush store.s
 
-  let close ?(flush = true) store =
+  let close ?(flush = true) ?(sync = true) store =
```
mark-christiaens, line 292: Maybe not use default values for flush and store.  This is the type of decision that you need to take very consciously to avoid mistakes.

src/node/store.ml
```diff
@@ -87,7 +87,7 @@ module type Simple_store = sig
   val delete_prefix: t -> transaction -> string -> int
 
   val flush: t -> unit Lwt.t
-  val close: t -> bool -> unit Lwt.t
+  val close: t -> flush:bool -> sync:bool -> unit Lwt.t
```
mark-christiaens, line 90: I don't understand the use of the "flush" parameter.  Is there a use case to close the store without flushing it?  Maybe when we need to do an "emergency" close without storing our internal batched operations to disk?  I think we should ensure that a sync guarantees a flush.  There's no point in syncing something that's not flushed.

domsj, line 90: In 1.6 there have been some issues regarding shutdown of arakoon nodes.
One of the things we tried is to no longer flush the store to disk so that shutdown can go as quick and smooth as possible (while remaining correct).

src/node/store.ml
domsj, line 292: In order to avoid mistakes I chose to have the most safe values as default .. but I could indeed make these arguments non-optional.

src/node/store.ml
mark-christiaens, line 90: Fine by me.  Ship it.

PR #452, opened 2014-06-18
Speedup batched store

NicolasT:


domsj:
Yes, LGTM!
Good find, apparently I was a bit stupid when I changed the batched store to the current mechanism ;-)

NicolasT:
I'll push it through CI first.

PR #454, opened 2014-06-18
1.6 merlin

mark-christiaens:


.merlin
```diff
@@ -0,0 +1,35 @@
...
+B _build/src/paxos
+B _build/src/plugins
+B _build/src/system
+B _build/src/tlog
+B _build/src/tools
+
+PKG bz2
+PKG camltc
+PKG lwt
+PKG lwt.ssl
```
NicolasT: Arakoon 1.6 doesn't depend on `lwt.ssl`, `quickcheck`, `snappy` and `ssl`.

mark-christiaens: Fixed

NicolasT:
LGTM

NicolasT:
Looks like I failed to spot a missing `B _build/src/node` line. Will fix.

PR #458, opened 2014-06-19
Add ability to list files for which warning 4 is disabled

NicolasT:
Since we'd like to enable the 'fragile pattern matches' warning, but
it's hard to get rid of it in some files (mostly pattern-matching on
`Unix.error`), we need a way to disable to corresponding warning for
these specific files.

This change adds the feature to `myocamlbuild.ml`, not whitelisting any
file for now.


toolslive:
The alternatives seem even more sub-awesome. LGTM.

NicolasT:
Once #448 is merged and forward-merged to 1.7 (with warning/error 4 enabled), the list can be populated selectively.

PR #461, opened 2014-06-24
Add TLS version & cipher list configuration

NicolasT:
These commits add the ability to select a TLS version to use (instead of the hard-coded 1.0), and configure a specific cipher list.

Note: as mentioned in the docstring (changed through one of the commits in this PR), the Python client can't handle any TLS version other than 1.0 due to Python 2.x restrictions.

TODO:
- [x] Extend documentation
- [x] Pass CI (running, `arakoon-git-launcher` 337)

### DIFF
--- a/src/node/node_cfg.ml
+++ b/src/node/node_cfg.ml
@@ -539,14 +539,17 @@ module Node_cfg = struct
     let get_string_option x = Ini.get inifile node_name x (Ini.p_option Ini.p_string) (Ini.default None) in
     let tls_cert = get_string_option "tls_cert"
     and tls_key = get_string_option "tls_key" in
-    if (tls_cert = None && tls_key <> None) || (tls_cert <> None && tls_key = None)
-      then failwith (Printf.sprintf "%s: both tls_cert and tls_key should be provided" node_name);
     let node_tls =
+      let msg = Printf.sprintf "%s: both tls_cert and tls_key should be provided" node_name in
       match tls_cert with
-        | None -> None
-        | Some cert -> match tls_key with
-            | None -> failwith "Node_cfg: Impossible!"
+        | None -> begin match tls_key with
+            | None -> None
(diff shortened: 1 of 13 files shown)


NicolasT:
`arakoon-git-base` 337 failed due to invalid `ssl` version. Running 338 after 2c94993.

NicolasT:
More CI issues... 339 running.

NicolasT:
`base` and `quick` passed as part of `arakoon-git-launcher` 339.

NicolasT:
CI (`arakoon-git-base` 339) passed, all blue.

src/node/node_cfg.ml
```diff
@@ -458,6 +541,13 @@ module Node_cfg = struct
     and tls_key = get_string_option "tls_key" in
     if (tls_cert = None && tls_key <> None) || (tls_cert <> None && tls_key = None)
       then failwith (Printf.sprintf "%s: both tls_cert and tls_key should be provided" node_name);
+    let node_tls =
+      match tls_cert with
+        | None -> None
+        | Some cert -> match tls_key with
+            | None -> failwith "Node_cfg: Impossible!"
```
domsj: What do you mean with 'impossible'?
That this code path is unreachable or is it a message for the user who did some misconfiguration?
(In the second case I think we might want a better message pointing the user in the correct direction...)

NicolasT: Unreachable due to check 2 lines above. Both `cert` and `key` are set, or none of them are set.

NicolasT: I guess I could move the `failwith` of above in here, and remove the check above.

domsj: Ah ok, somehow I missed that check (apparently I was only looking at changed code).
What you suggest could indeed be clearer

domsj:
LGTM

PR #462, opened 2014-06-24
Logging

mark-christiaens:


domsj:
Except for my remark about `if true then` this looks ok to me

src/node/statistics.ml
```diff
@@ -197,9 +199,16 @@ module Statistics = struct
     update_x_stats t.op_time_stats x ;
     x
 
+  let log_slowness descr total_time =
+    if total_time > 1.0 then
+    if true then
```
NicolasT: This is bizarre.

mark-christiaens: Slip of the finger.  Fixed.

PR #463, opened 2014-06-24
ensure tlfs are synced before renaming them to their destination

domsj:


src/tools/file_system.ml
```diff
@@ -92,6 +92,20 @@ let exists filename =
       | e -> Lwt.fail e
     )
 
+let with_output_file tmp f =
```
NicolasT: I'd keep the 'open flags' & permission mask explicit as arguments.

domsj: Dunno .. I thought these should always have this value (for the semantics of this function); so why think about it at every call site (and introduce possible inconsistencies there)?
I could make them optional arguments with the current values as defaults ... but why make something more generic than needed?

NicolasT: What if you want `O_APPEND`, or don't care about `O_EXCL`?

domsj: What if I rename the function to with_new_exclusive_output_file? (Or maybe a with_temp_file ... which also handles the renaming etc ...)
For the semantics I have in mind I don't want O_APPEND or O_EXCL.
(And again, why make it more generic than needed?)

NicolasT: I like the `with_temp_file` idea which also performs renaming (and maybe unlinking of the temporary file beforehand).

domsj:
Pushed another commit, now rerunning CI ( http://172.19.49.85/view/arakoon-git/job/arakoon-git-launcher/351/ )

domsj:
Hmm, there seem to be some CI failures now which I should investigate

domsj:
Fixed the issues, CI is happy again

NicolasT:
LGTM

PR #464, opened 2014-06-24
Add & use `to_string` utility API

NicolasT:


toolslive:
it's more convenient (and faster) if you work with an intermediate sequence type. Maybe you should seek inspiration in the Wadler/Leijen prettier printer.

NicolasT:
I'm certainly familiar with the Wadler & Leijen approach to pretty-printers. The use-case here isn't generic documents though. I think creating some Builder-like monoidal system and some interpreter for it is overkill, and adds more complexity to the user without gain: it's not the intention to be able to create/design arbitrary documents!

domsj:
Agreed that we shouldn't make this unnecessarily generic.
Except for the leading `;`s this LGTM ;-)

PR #467, opened 2014-06-26
Update `arakoon.odocl`

NicolasT:
Some modules were removed lately, some were added,... Updating the
documented module list.


domsj:
Ok

PR #468, opened 2014-06-26
Add Ocamldoc comments for some library modules

NicolasT:
Depends on #467.

### DIFF
--- a/src/tools/ring_buffer.ml
+++ b/src/tools/ring_buffer.ml
@@ -14,50 +14,48 @@
  * limitations under the License.
  *)
 
+(** A simple ring-buffer implementation *)
+
+(** Create a new ring-buffer for a given element type *)
 module Make(E : sig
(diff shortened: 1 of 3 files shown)


src/tools/ring_buffer.ml
```diff
@@ -14,50 +14,48 @@
...
              }
 
+    (** Create a new ring-buffer of given size
+
+    @param size Size of the ring-buffer *)
     let create ~size =
         let buffer = Array.create size E.zero in
         { buffer; pos = 0; }
 
+    (** Insert (in-place) a new element in the ring-buffer *)
```
domsj: Not sure about the phrasing here, would `Insert an element at the next position in the ring-buffer` be better?
(What's `new` about the element being inserted?)

domsj:
I suppose you 'built' this new documentation. Minor nit but looks ok to me

PR #470, opened 2014-06-26
Multiple masters

domsj:
Running CI a second time just to be sure, but looking good so far.

### DIFF
--- a/src/paxos/master.ml
+++ b/src/paxos/master.ml
@@ -49,7 +49,9 @@ let master_consensus constants {mo;v;n;i;lew} () =
   )
   in
   let state = (v,n,(Sn.succ i), lew) in
-  Fsm.return ~sides:[con_e;log_e;inject_e] (Stable_master state)
+  if Value.is_other_master_set constants.me v
+  then Fsm.return ~sides:[con_e;log_e;] (Slave_wait_for_accept (n,Sn.succ i,None))
+  else Fsm.return ~sides:[con_e;log_e;inject_e] (Stable_master state)
 
 
 let null = function
@@ -173,7 +175,7 @@ let stable_master (type s) constants ((v',n,new_i, lease_expire_waiters) as curr
 		              begin
                         let l_val = constants.tlog_coll # get_last () in
                         Multi_paxos.safe_wakeup_all () lease_expire_waiters >>= fun () ->
-			            Fsm.return (Slave_wait_for_accept (n', new_i, None, l_val))
+			            Fsm.return (Slave_wait_for_accept (n', new_i, l_val))
 		              end
 		            | Promise_sent_needs_catchup ->
                       let i = S.get_catchup_start_i constants.store in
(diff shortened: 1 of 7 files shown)


domsj:
Second run looks good too

src/paxos/value.ml
```diff
@@ -33,7 +33,12 @@ let is_master_set  = function
   | Vm _ ->  true
   | _    -> false
 
-let is_synced = function 
+let is_other_master_set me = function
+  | Vm (m, _) when m = me -> false
```
NicolasT: I'd prefer

``` ocaml
let is_other_master_set me = function
  | Vm (m, _) -> m <> me
  | Vc _ -> false
```

instead of guards.

src/paxos/master.ml
```diff
@@ -49,7 +49,9 @@ let master_consensus constants {mo;v;n;i;lew} () =
   )
   in
   let state = (v,n,(Sn.succ i), lew) in
-  Fsm.return ~sides:[con_e;log_e;inject_e] (Stable_master state)
+  if Value.is_other_master_set constants.me v
+  then Fsm.return ~sides:[con_e;log_e;] (Slave_wait_for_accept (n,Sn.succ i,None))
```
NicolasT: For clarity, I'd consider writing something like

``` ocaml
let stepdown ~sides ~n ~i ~some_descriptive_name =
  let sides = (sides @ [ELog (fun () -> "Stepping down to slave state")]) in (* Assuming `sides` are interpreted in-order *)
  Fsm.return ~sides (Slave_wait_for_accept (n, i, some_descriptive_name))
```

higher up in the module, and then obviously

``` ocaml
then stepdown ~sides ~n ~i:(Sn.succ i) ~some_descriptive_name:None
```

NicolasT:
The intent LGTM, but maybe this is a really good one for @mark-christiaens, @lienboe or @liegebeest to review, reason about & validate as well :smiley:

src/paxos/master.ml
```diff
@@ -48,8 +48,14 @@ let master_consensus constants {mo;v;n;i;lew} () =
     | _ -> Lwt.return ()
   )
   in
-  let state = (v,n,(Sn.succ i), lew) in
-  Fsm.return ~sides:[con_e;log_e;inject_e] (Stable_master state)
+  if Value.is_other_master_set constants.me v
+  then
+    (* step down *)
+    let sides = [con_e;log_e;ELog (fun () -> "Stepping down to slave state")] in
+    Fsm.return ~sides (Slave_wait_for_accept (n, Sn.succ i, None)) in
```
NicolasT: Oops?

domsj: Why the oops ?

I think I did forget to `Multi_paxos.safe_wakeup_all () lease_expire_waiters` here ; will adjust.

domsj: Ah right, now I see the oops.

PR #471, opened 2014-06-27
Rework `Multi_paxos.update_votes`

NicolasT:
Since it wasn't immediately obvious to me what `Multi_paxos.update_votes` did while reviewing #470, this change attempts to simplify things by introducing a fairly generic `Std.List.alter` function, which is then used to implement `Multi_paxos.update_votes`, hopefully clarifying what this function does exactly.
- [x] Run CI


NicolasT:
`arakoon-git-base` 348 all blue.

domsj:
LGTM if CI is ok

NicolasT:
`arakoon-git-launcher 399` launched because of merge.

NicolasT:
All blue.

PR #474, opened 2014-07-01
at end of collapse, replaced copy with a rename.

toolslive:


NicolasT:
Maybe this is an opportunity to remove the `.clone` file as well in the `finalize` block, if it still exists?

src/node/collapser.ml
```diff
@@ -118,37 +118,51 @@ let collapse_until (type s) (tlog_coll:Tlogcollection.tlog_collection)
...
+      (fun () -> File_system.exists new_location >>=
+                   function
+                   | true ->
+                      begin
+                        Logger.info_f_ "unlinking %s" new_location >>= fun ()->
+                        File_system.unlink new_location
+                      end
+                   | false -> Lwt.return ())
+      (fun exn ->
+       Logger.warning_f_ "ignoring failure in cleanup: %S"
```
NicolasT: `warning_f_` adds the exception to the logged message when `~exn` is passed along:

``` ocaml
let section = Lwt_log.Section.main

let main () =
    let exn = Failure "Fail!" in
    Logger.warning_f_ ~exn "Foo"
;;

Lwt_main.run (main ())
```

```
$ ./test_logger.native 
test_logger.native: main: Foo: Failure("Fail!")
```

NicolasT:
Cava.

PR #479, opened 2014-07-07
Add `Client_main.find_master'`

NicolasT:
This patch adds `Client_main.find_master'`, which should be a more resilient master lookup algorithm.

It also reworks `Client_main.find_master` to use the new `find_master'`, but keep the behaviour of the original action regarding error handling.

### DIFF
--- a/src/client/client_main.ml
+++ b/src/client/client_main.ml
@@ -75,11 +75,11 @@ let with_connection ~tls sa do_it = match tls with
 
 exception In_progress
 exception Done
-exception Unknown_error
+exception Impossible
 
 let with_connection' ~tls addrs f =
-  let m = ref None
-  and e = Lwt_mvar.create None
+  let result = Lwt_mvar.create None
+  and error = Lwt_mvar.create None
   and l = Lwt_mutex.create () in
 
   let f' addr =
@@ -90,26 +90,36 @@ let with_connection' ~tls addrs f =
             then Lwt.fail In_progress
             else begin
               Lwt_mutex.with_lock l (fun () ->
-              match !m with
-                | Some _ -> Lwt.fail Done
-                | None -> begin
-                    f addr c >>= fun v ->
-                    m := Some v;
-                    Lwt.return ()
-                end)
+                Lwt_mvar.take result >>= function
+                  | Some _ -> Lwt.fail Done
+                  | None -> begin
+                      Lwt.catch
+                        (fun () ->
+                          f addr c >>= fun v ->
+                          Lwt_mvar.put result (Some v))
+                        (fun exn ->
+                          Lwt_mvar.put result None >>= fun () ->
+                          Lwt_mvar.take error >>= fun _ ->
+                          Lwt_mvar.put error (Some exn))
+                  end)
             end))
-      (fun exn ->
-        Lwt_mvar.take e >>= function
-          | Some _ as v -> Lwt_mvar.put e v
-          | None -> Lwt_mvar.put e (Some exn))
+      (function
+        | In_progress -> Lwt.return ()
+        | Done -> Lwt.return ()
+        | exn -> begin
+            Lwt_mvar.take error >>= begin function
+              | Some _ as v -> Lwt_mvar.put error v
+              | None -> Lwt_mvar.put error (Some exn)
+            end
+        end)
   in
 
   let ts = List.map f' addrs in
   Lwt.join ts >>= fun () ->
-  match !m with
+  Lwt_mvar.take result >>= function
     | Some v -> Lwt.return v
-    | None -> Lwt_mvar.take e >>= function
-        | None -> Lwt.fail Unknown_error
+    | None -> Lwt_mvar.take error >>= function
+        | None -> Lwt.fail Impossible
         | Some exn -> Lwt.fail exn
 
 
--- a/src/client/client_main.ml
+++ b/src/client/client_main.ml
@@ -76,7 +76,16 @@ let with_connection ~tls sa do_it = match tls with
 exception Impossible of string
 
 let with_connection' ~tls addrs f =
-  let result = Lwt_mvar.create None
+  (* Final result:
+   *
+   * - `Empty: No connection succeeded
+   * - `Success of 'a: One connection succeeded and `f` returned a value
+   * - `Failure of exn: One connection succeeded and `f` raised an exception
+   *)
+  let result = Lwt_mvar.create `Empty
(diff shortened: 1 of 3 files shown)


NicolasT:
- [x] CI: `arakoon-git-launcher` 356
- [x] Attempt to connect to all IP addresses of a node concurrently?

domsj:
I think we should definitely try connecting on all ips of a node, and while it could be done concurrently doing it sequentially would already be an important improvement too (and actually be good enough for me).

Anyway, this definitely looks like an improvement so far.

NicolasT:
Implemented concurrent connecting.

NicolasT:
- [ ] `arakoon-git-launcher` 361

src/client/client_main.ml
```diff
@@ -69,24 +69,68 @@ let with_connection ~tls sa do_it = match tls with
...
-      (fun exc ->
-        Lwt_ssl.close sock >>= fun () ->
-        Lwt.fail exc)
+        with_connection ~tls addr (fun c ->
+          if Lwt_mutex.is_locked l
+            then Lwt.fail In_progress
+            else begin
+              Lwt_mutex.with_lock l (fun () ->
+                Lwt_mvar.take result >>= function
+                  | Some _ -> Lwt.fail Done
```
domsj: - this can't actually happen
- if it did happen you would want to put the result back into the mvar

src/client/client_main.ml
```diff
@@ -69,24 +69,68 @@ let with_connection ~tls sa do_it = match tls with
...
       (fun () ->
-        do_it (ic, oc) >>= fun r ->
-        Lwt_ssl.close sock >>= fun () ->
-        Lwt.return r)
-      (fun exc ->
-        Lwt_ssl.close sock >>= fun () ->
-        Lwt.fail exc)
+        with_connection ~tls addr (fun c ->
+          if Lwt_mutex.is_locked l
+            then Lwt.fail In_progress
```
domsj: Why not just `Lwt.return_unit`?

src/client/client_main.ml
```diff
@@ -69,24 +69,68 @@ let with_connection ~tls sa do_it = match tls with
     Typed_ssl.Lwt.ssl_connect fd ctx >>= fun (_, sock) ->
     let ic = Lwt_ssl.in_channel_of_descr sock
     and oc = Lwt_ssl.out_channel_of_descr sock in
+    Lwt.finalize
+      (fun () -> do_it (ic, oc))
+      (fun () -> Lwt_ssl.close sock)
+
+exception In_progress
+exception Done
+exception Impossible
```
domsj: I don't think you need any of these...

src/client/client_main.ml
```diff
@@ -69,24 +69,109 @@ let with_connection ~tls sa do_it = match tls with
...
-  let sa = _address_to_use node_cfg.ips node_cfg.client_port in
-  let do_it connection =
+  let addrs = List.map (fun ip -> make_address ip node_cfg.client_port) node_cfg.ips in
+  let do_it _ connection =
     Arakoon_remote_client.make_remote_client cluster connection
     >>= fun (client: Arakoon_client.client) ->
     f client
   in
-  with_connection ~tls sa do_it
+  with_connection' ~tls addrs do_it
```
NicolasT, line 139: This can raise other exceptions than the original implementation (`No_connection` instead of some `Unix_error`). Is that a problem?

src/client/client_main.ml
```diff
@@ -69,24 +69,109 @@ let with_connection ~tls sa do_it = match tls with
...
+
+
+module CountDownLatch : sig
+    type t
+
+    val create : count:int -> t
+    val await : t -> unit Lwt.t
+    val count_down : t -> unit Lwt.t
+end = struct
+    type t = { mvar : int Lwt_mvar.t
```
domsj: I don't understand why this is an mvar instead of a `int ref` ?
I think we could drop the Lwt_mutex.t below here too

src/client/client_main.ml
```diff
@@ -69,24 +69,109 @@ let with_connection ~tls sa do_it = match tls with
...
+          then Lwt.return ()
+          else
+          Lwt_mutex.lock l >>= fun () ->
+            Lwt.catch
+              (fun () -> f addr c >>= fun v -> Lwt_mvar.put res (`Success v))
+              (fun exn -> Lwt_mvar.put res (`Failure exn))))
+      (fun exn ->
+        CountDownLatch.count_down cd >>= fun () ->
+        Logger.warning_f_ ~exn "Failed to connect" >>= fun () ->
+        Lwt.fail exn)
```
domsj, line 107: Hmm, I expected all exceptions to be caught here, saved into some refcell, and then possible rethrown where you throw the No_connection now?
(That way with_connection would still throw the same exceptions, to answer your comment a bit below here)

src/tools/lwt_extra.ml
```diff
@@ -0,0 +1,73 @@
...
+                  let mutex = t.mutex in
+                  Lwt_condition.wait ~mutex t.condition
+              end)
+
+    (** Decrement the latch by one *)
+    let count_down t =
+        Lwt_mutex.with_lock t.mutex (fun () ->
+            Lwt.protected (
+                Lwt_mvar.take t.mvar >>= fun c ->
+                let c' = max (c - 1) 0 in
```
domsj, line 67: Don't we want to throw an exception here when the count would go below 0?
Seems like an incorrect usage to me... (but I honestly haven't looked at APIs for countdown latches in other places, so not sure what the standard behavior there is)

NicolasT, line 67: Not necessarily.

Say you got a networked system, consisting of 5 target nodes. In some algorithm, you want to send some message (and receive an acknowledgement) to at least 3 of these nodes, concurrently. Then you use a `CountDownLatch` of size 3, spawn 5 threads which send the message & wait for an answer, and once the ack has been received, decrease the latch.

Finally the caller blocks on the latch (& some timeout, or whatever).

It's perfectly possible and OK if the latch is decremented even though it's 0 already when more-than-3 messages are sent/ack'ed.

domsj, line 67: Ok, sounds reasonable.
(Fwiw, apparently in Java it has the same semantics.)

src/client/client_main.ml
NicolasT, line 139: Fixed.

domsj:
LG if CI is ok

PR #480, opened 2014-07-08
Preferred master 1.7

domsj:
- [x] waiting for CI http://172.19.49.85/view/arakoon-git/job/arakoon-git-launcher/358/downstreambuildview/?

After that we probably want to do another backport...

### DIFF
--- a/pylabs/test/server/right/system_tests_preferred_masters.py
+++ b/pylabs/test/server/right/system_tests_preferred_masters.py
@@ -59,7 +59,6 @@ def test_preferred_master():
 
     # node should no longer be preferred after the dropMaster call
     Common.dropMaster(pm)
-    time.sleep(Common.lease_duration)
     client.nop()
     master4 = client.whoMaster()
     NT.assert_not_equals(master4, pm)
--- a/pylabs/test/server/right/system_tests_preferred_masters.py
+++ b/pylabs/test/server/right/system_tests_preferred_masters.py
@@ -64,9 +64,8 @@ def test_preferred_master():
     master4 = client.whoMaster()
     NT.assert_not_equals(master4, pm)
 
-    # restart node to make it behave in a preferred manner again
-    cluster.stopOne(pm)
-    cluster.startOne(pm)
+    # perform maintenance on node to make it behave in a preferred manner again
+    Common.defragDb(pm)
 
     time.sleep(Common.lease_duration)
     client.nop()
--- a/pylabs/test/server/right/system_tests_preferred_masters.py
+++ b/pylabs/test/server/right/system_tests_preferred_masters.py
@@ -67,7 +67,7 @@ def test_preferred_master():
     # perform maintenance on node to make it behave in a preferred manner again
     Common.defragDb(pm)
 
-    time.sleep(Common.lease_duration)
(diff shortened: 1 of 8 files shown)


domsj:
With this implementation I think it should also be easy to have the preferred node start acting in a preferred manner again after e.g. 15 minutes, which might be desirable behavior...

domsj:
Looks good on CI. I have not implemented the idea I mentioned above due to lack of feedback, can still be done later.

src/client/client_main.ml
```diff
@@ -19,52 +19,25 @@ open Network
...
-             -> protocol : Ssl.protocol
-             -> t
-
-    val ca_cert : t -> path
-    val creds : t -> (path * path) option
-    val protocol : t -> Ssl.protocol
-end = struct
-    type path = string
+let default_initialize_client_context ~ca_cert ~creds ctx =
+  Typed_ssl.set_verify ctx
```
NicolasT, line 29: Hmh, not sure this is a good idea... Instead of passing in a context, shouldn't this create & return one? I don't think it's the intention to do this multiple times on the same context.

pylabs/test/server/right/system_tests_preferred_masters.py
```diff
@@ -57,6 +57,21 @@ def test_preferred_master():
     logging.info("master3 = %s", master3)
     NT.assert_equals(master3,pm)
 
+    # node should no longer be preferred after the dropMaster call
+    Common.dropMaster(pm)
+    time.sleep(Common.lease_duration)
```
NicolasT: Skype comments.

src/client/client_main.ml
```diff
@@ -129,15 +102,24 @@ let with_connection' ~tls addrs f =
...
+    Arakoon_remote_client.make_remote_client cluster_id connection
     >>= fun (client: Arakoon_client.client) ->
     f client
   in
   with_connection' ~tls addrs do_it
 
+let with_remote_nodestream ~tls node_cfg cluster_id f =
+  let addrs = List.map (fun ip -> make_address ip node_cfg.client_port) node_cfg.ips in
+  let do_it _addr connection =
+    Remote_nodestream.make_remote_nodestream cluster_id connection
```
NicolasT:
``` ocaml
Remote_nodestream.make_remote_nodestream cluster_id connection >>= f
```

src/client/remote_nodestream.ml
```diff
@@ -182,8 +182,12 @@ class remote_nodestream ((ic,oc) as conn) =
 end :nodestream)
 
 
-let make_remote_nodestream cluster connection =
-  prologue cluster connection >>= fun () ->
+let make_remote_nodestream ?(skip_prologue=false) cluster connection =
+  begin
+    if skip_prologue
```
NicolasT, line 187: I should add my _monad utils Functor_ one day, so this coule be

```
unless skip_prologue 
    (prologue cluster connection) >>= fun () -> ...
```

https://github.com/NicolasT/ocaml-free/blob/d9533271249e19a7c5fee128e53fbbe59f6ceeb8/src/free_monad.ml#L203

NicolasT:
Looks OK if CI is happy, but as far as I understood it was.

PR #481, opened 2014-07-08
make closing a tlog with the correct marker a bit easier

domsj:


src/tlog/tlc2.ml
```diff
@@ -213,8 +213,8 @@ let fold_read tlog_dir tlf_dir file_name
 
 type validation_result = (Entry.t option * Index.index)
 
-let _make_close_marker node_id = Some ("closed:" ^ node_id)
-let _make_open_marker node_id =  Some ("opened:" ^ node_id)
+let _make_close_marker node_id = "closed:" ^ node_id
+let _make_open_marker node_id =  "opened:" ^ node_id
```
NicolasT, line 217: No need to change the 'opened' path I think?

domsj, line 217: Not really, but I wanted to keep it consistent.
The way I currently look at it a marker is a string, and a tlog entry contains a 'marker option'.

Hmm, instead of making the change like this I could've changed `Tlog_main.mark_tlog` to receive a string option as the new marker instead, maybe that would be better?

NicolasT, line 217: Hmh, no, because what should happen then when `None` is returned?

Anyway, I guess returning a `string` makes more sense than returning a `string option` which is always `Some ...`, so I retract my comment.

NicolasT:
Maybe consider changing or duplicating `system_tests_basic:test_marker_presence_required` to use this mechanism so it's 'tested' as well?

domsj:
I've expanded `system_tests_basic:test_marker_presence_required` a bit to test this mechanism. (Still waiting on CI)

Do you think the extra `-close` is clear enough? I could instead make a real separate call `--close-tlog`...

NicolasT:
I was thinking about that yesterday as well.Having 2 top-level actions might be confusing as well though?

NicolasT:
OK if CI is happy.

PR #482, opened 2014-07-11
simple throttling scheme for head copying

toolslive:
crossref: [http://jira.incubaid.com/browse/ARAKOON-476]


src/node/node_cfg.ml
```diff
@@ -381,7 +383,13 @@ module Node_cfg = struct
       try get_string "log_dir"
       with _ -> home
     in
-    let reporting = Ini.get inifile node_name "reporting" Ini.p_int (Ini.default 300) in
+    let reporting = Ini.get inifile node_name "reporting"
+                            Ini.p_int (Ini.default 300)
+    in
+    let head_copy_throttling =
+      Ini.get inifile node_name "head_copy_throttling"
+              Ini.p_float (Ini.default 0.0)
```
domsj: I think you added `default_head_copy_throttling` above so you could use it here

src/node/node_main.ml
```diff
@@ -435,7 +435,9 @@ let _main_2 (type s)
           let full_snapshot_path = Filename.concat me.head_dir snapshot_name in
           Lwt.catch
             (fun () ->
-              S.copy_store2 full_snapshot_path db_name false
+              S.copy_store2 full_snapshot_path db_name
+                            ~overwrite:false
+                            ~throttling:me.head_copy_throttling
```
domsj: Hmm, from the jira ticket I assumed we only wanted to do the throttling when copying the head in the first step of collapsing?
Now I'm a bit confused about which places we want to do throttling...

src/tools/file_system.ml
```diff
@@ -78,17 +78,28 @@ let exists filename =
       | e -> Lwt.fail e
     )
 
-let copy_file source target ~overwrite = (* LOOKS LIKE Clone.copy_stream ... *)
+let copy_file source target ~overwrite ~throttling =
+  (* LOOKS LIKE Clone.copy_stream ... *)
   Logger.info_f_ "copy_file %s %s" source target >>= fun () ->
```
domsj: I would log `(throttling:%f)` here instead of in collapser.ml

src/tools/file_system.ml
```diff
@@ -78,17 +78,28 @@ let exists filename =
...
   let bs = Lwt_io.default_buffer_size () in
   let buffer = String.create bs in
+  let throttle =
+    match throttling with
+    | 0.0   -> fun _  -> Lwt.return ()
+    | f -> fun dt -> Lwt_unix.sleep (f *. dt)
+  in
   let copy_all ic oc =
     let rec loop () =
+      let t0 = Unix.gettimeofday() in
```
NicolasT: This adds 2 `Unix.gettimeofday ()` calls, even when throttling is disabled. Consider using a wrapper which is only used when throttling is enabled?

domsj:
Handled the review comments.
Giving it another CI run http://cimaster.incubaid.com/view/arakoon-git/job/arakoon-git-launcher/368/

NicolasT:
LGTM if CI is happy.

PR #483, opened 2014-07-11
ensure a node sleeps a lease period on startup if another node was previ...

domsj:
...ously the master

(this backports some changes from the 1.7 branch)


NicolasT:
This is a backport of 1.7, right? Seems OK if CI is happy.

PR #485, opened 2014-07-17
return an exit code different from 0 when some tests fail

domsj:


src/main/main.ml
```diff
@@ -123,19 +134,19 @@ let run_some_tests repeat_count filter =
...
-  let rec do_n_times f = function
-    | 0 -> ()
+  let rec do_n_times f results = function
+    | 0 -> results
     | n ->
       let () = show_count n in
-      let (_:OUnit.test_result list) = f () in
-      do_n_times f (n-1)
+      let results' = f () in
+      do_n_times f (List.append results' results) (n-1)
```
NicolasT, line 142: Wouldn't it make more sense to append the other way around so test results are kept in execution order?

NicolasT:
Seems OK if interaction with CI is as desired.

PR #486, opened 2014-07-22
backport preferred master changes from 1.7 + required dependencies

domsj:
(GET_TXID, NOP, Client_main.with_connection')

Gonna run it through CI again when it's back up & running, but you can already have a look.


NicolasT:
OK if CI is happy.

Although I think this is one more reason why one should just move from 1.6 to 1.7.

PR #487, opened 2014-07-22
add database corruption warning to --close-tlog description

domsj:
Had to explain this to a user today.
Trying to keep the universe from winning...

### DIFF
--- a/src/main/main.ml
+++ b/src/main/main.ml
@@ -246,11 +246,11 @@ let main () =
                                Arg.Set_string key;
                              ],
      "<filename> <key>: add a marker to a tlog");
-    ("--close-tlog", Arg.Tuple[ set_laction CloseTlog;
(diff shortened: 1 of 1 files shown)


src/main/main.ml
```diff
@@ -250,7 +250,7 @@ let main () =
                                Arg.Set_string filename;
                                Arg.Set_string node_id;
                              ],
-     "<filename> <node_name>: marks the tlog with 'closed:node_name'");
+     "<filename> <node_name>: marks the tlog with 'closed:node_name'. Please note that if you require doing this the tokyo cabinet database might be silently CORRUPTED.");
```
NicolasT: I'd slightly rephrase this, along the lines of `Note when a tlog marker is missing, the node Tokyo Cabinet database could be silently corrupted, which is not fixed by this command.`

NicolasT: Rename to `--unsafe-close-tlog` ?

NicolasT:
Cava

PR #488, opened 2014-07-24
1.6 for 1.7

domsj:
Can't check CI in detail any more...


NicolasT:
I restarted TravisCI a couple of times, and several of the base tests keep failing with `No master` etc. :-/

PR #489, opened 2014-08-01
limits dependencies for plugins

toolslive:
Plugins need to be able to (de)serialize inputs and outputs, but allowing them to access `Llio` opens up `Std` and explodes the number of dependencies. These dependencies make _plugins_ break if they are not compiled with the exact (md5 wise) same set of library versions as the arakoon binary.

The patch introduces an extra interface allowing them to not be dependent on kitchen sinks like `Lwt` or `Std`. 
As an example: 

```
ocamlobjinfo _build/plugin_find.cmo
File _build/plugin_find.cmo
Unit name: Plugin_find
Interfaces imported:
    ad06f04cfca6d404d1de76c3dc67324a    Int32
    d012329cc712e91d0f10a5eef2303d18    Printf
    db7f34081ef8fcaf499f19523d0736c6    String
    9b043210ba766ec98c0c64e3dfba3013    Plugin_find
    e2378c62e62b3516b3b9da1a984c728d    Plugin_helper
    af3ef6fba94cdb4eba31e98b4e341dab    Buffer
    024edc3512403b725052aec8e41ed971    Hashtbl
    b0adfa4175f86e4394859886c1a374bb    Obj
    36b5bc8227dc9914c6d9fd9bdcfadb45    Pervasives
    f11d7ddcffad09397202a49bd9bb4283    Map
    6c35bd33088907f850bc27cd9072ffb5    Registry
Uses unsafe features: no
Force link: no
```

While previously it would fe also depend on `Test_utils`, `Quickcheck` and some 40 others.

### DIFF
--- a/src/plugins/plugin_helper.ml
+++ b/src/plugins/plugin_helper.ml
@@ -1,0 +1,16 @@
+let serialize_string  b s = Llio.string_to b s
+let serialize_hashtbl b f h = Llio.hashtbl_to b f h
+let serialize_string_list b sl = Llio.string_list_to b sl
+
+type input = Llio.buffer
+let make_input s i = Llio.make_buffer s i
+
+let deserialize_string i = Llio.string_from i
+let deserialize_string_list i = Llio.string_list_from i
+let deserialize_hashtbl i f = Llio.hashtbl_from i f
+
+let generic_f f x =
+  let k s = Lwt.ignore_result (f s) in
+  Printf.kprintf k x
+
+let debug_f x = generic_f Client_log.debug x
--- a/src/plugins/plugin_helper.ml
+++ b/src/plugins/plugin_helper.ml
@@ -1,3 +1,19 @@
+(*
+Copyright (2010-2014) INCUBAID BVBA
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*)
+
 let serialize_string  b s = Llio.string_to b s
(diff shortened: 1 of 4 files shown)


NicolasT:
New modules lack a license header.

src/plugins/plugin_loader.ml
```diff
@@ -19,6 +19,14 @@ open Lwt
 let section = Logger.Section.main
 
 let load home pnames =
+  (*let () = Dynlink.allow_only [
```
NicolasT: Should this be removed, or uncommented & extended?

src/plugins/plugin_loader.ml
```diff
@@ -27,7 +35,16 @@ let load home pnames =
...
+        try
+          let () = Dynlink.loadfile_private qual in
+          None
+        with Dynlink.Error e ->
+          let em = Dynlink.error_message e in
+          Some em
+      in
+      match msg with
+      | None -> _inner ps
+      | Some em -> failwith (Printf.sprintf "%s: Dynlink.Error %S" p em)
```
NicolasT, line 53: Consider failing with some specific exception type?

src/plugins/plugin_loader.ml
toolslive: none of the above. 
it serves as a hint for if somebody would like to do this.

src/plugins/plugin_loader.ml
toolslive, line 53: what's the use? we're just interested in going down.

src/plugins/plugin_helper.ml
```diff
@@ -0,0 +1,32 @@
...
+
+let deserialize_string i = Llio.string_from i
+let deserialize_string_list i = Llio.string_list_from i
+let deserialize_hashtbl i f = Llio.hashtbl_from i f
+
+let generic_f f x =
+  let k s = Lwt.ignore_result (f s) in
+  Printf.kprintf k x
+
+let debug_f x = generic_f Client_log.debug x
```
domsj, line 32: Do we want to log this to a specific section (e.g. 'plugins') ?

src/plugins/plugin_loader.ml
domsj: Maybe mention this is a hint for future implementors (and when/why they would want to do this), otherwise somebody might come and clean it up

PR #490, opened 2014-08-11
Checksums

314eter:


DESIGN.md
```diff
@@ -0,0 +1,72 @@
...
+
+### New value format
+* 0x100 (int32)
+* Checksum (int32 if crc-32 is used)
+* Value type (char ‘c’ or ‘m’)
+* Value details (depends on type)
+
+Upgrade Path
+------------
+To upgrade Arakoon to the new version, all nodes need to be restarted simultaneously. The tlogs in the old format can be read, and the checksums of these entries will be set to None (Checksum is an option type). Values with checksum None will not be valuated.
```
domsj: Restarting all nodes simultaneously in a distributed system can be a bit hard, and leads to longer downtime.
I think it will probably be easier to restart the nodes 1 by 1. At a certain point the majority switches from the old nodes to the new ones.
You probably want to keep things easy by not having old and new nodes communicate with each other, right?
This can be done by either
- changing the port of the inter node communication or
- by having some magic mismatch when a connection is made between old and new nodes (I'm not sure if there currently is a magic ... this option would probably protect better against operator errors, so therefor I would prefer it)

src/node/node_main.ml
```diff
@@ -895,6 +895,11 @@ let _main_2 (type s)
             let rc = 50 in
             Logger.fatal_f_ "[rc=%i] Somebody has been messing with the available tlogs" rc >>= fun () ->
             Lwt.return rc
+          | Value.ValueCheckSumError (i, value) ->
```
domsj, line 906: I might still be missing something, but what's the difference between this error and the Tlogcommon.TLogCheckSumError a few lines above?

NicolasT, line 906: Whilst `ValueChecksumError` might not be the best name, I think it makes sense not to use `TLogCheckSumError`. The latter is used to denote on-disk issues: an invalid checksum found in a TLog file, denoting local TLog corruption.

This issue can (with caution, and violating Paxos requirements) be fixed by truncating a TLog.

On the other hand, when a conflict is discovered in the 'rolling' checksum, this doesn't denote an invalid local TLog, rather a local TLog which somehow diverged from the cluster state. Node-local there's nothing really 'wrong', and the case points at a major cluster-wide problem.

DESIGN.md
```diff
@@ -0,0 +1,72 @@
+Checksums in Arakoon
```
NicolasT: This file should get another name and moved into a subdirectory (most likely `doc`?).

DESIGN.md
NicolasT: Bumping the magic seems like a good idea. Otherwise nodes could get stuck trying to read unexpected data from the network. Note this only applies to the inter-node messaging magic, no reason to change the client protocol.

src/node/node_main.ml
```diff
@@ -895,6 +895,11 @@ let _main_2 (type s)
...
+            let rc = 51 in
+            Logger.fatal_f_ "[rc=%i] Value has a checksum error: sn=%s, value=%s"
```
NicolasT: It might be beneficial to add some more information in the message regarding which checksum it is all about.

src/paxos/value.ml
```diff
@@ -18,84 +18,142 @@ limitations under the License.
...
+      Llio.list_to buf Update.to_buffer us
+    end
+  | Vm (m,l) -> begin
+      Llio.char_to buf 'm';
+      Llio.string_to buf m;
+      Llio.int64_to buf (Int64.of_float l)
+    end
+
+let value_to buf (cs, c) = 
+  if Checksum.is_none cs
```
NicolasT: Looking at the code below, I guess there's some special 'empty' checksum value. Any reason not to use an `option` instead?

314eter: The checksum is an (abstract) option type. But instead of saving it in the tlog as an option, values with checksum None are saved in the old format.

src/tools/checksum.ml
```diff
@@ -0,0 +1,91 @@
...
+  val checksum_from : Llio.buffer -> t
+
+  val calculate : string -> t
+  val update : t -> string -> t
+end
+
+module type C = sig
+  include ChecksumType
+
+  val none : t
```
NicolasT: I'd just expose the `t` type to be `Cs.t option`, so any existing `option`-related utility function can work on it, no need for these special cases.

src/tools/checksum.ml
```diff
@@ -0,0 +1,91 @@
...
+module Make (Cs : ChecksumType) : C = struct
+  type t = Cs.t option
+
+  let none = None
+
+  let is_none = function
+    | None -> true
+    | Some _ -> false
+
+  let string_of = function
```
NicolasT: Take a look at the `To_string` module on how to write these things in a more uniform way.

src/tools/checksum.ml
```diff
@@ -0,0 +1,91 @@
...
+    | None -> "_"
+    | Some cs -> Cs.string_of cs
+
+  let of_string = function
+    | "_" -> None
```
NicolasT: Is there any reason to use `_` instead of `None` and `Some ...`?

src/tools/checksum.ml
```diff
@@ -0,0 +1,91 @@
...
+    | s -> Some (Cs.of_string s)
+
+  let checksum_to buf = function
+    | Some cs -> Cs.checksum_to buf cs
+    | None -> failwith "should not happen"
```
NicolasT: This introduces partiality. Couldn't the function take a `Cs.t` instead of a `t`, leaving the proof of its existence to the caller?

src/tools/checksum.ml
```diff
@@ -0,0 +1,91 @@
...
+  let string_of = Printf.sprintf "%lx"
+  let of_string s = Scanf.sscanf s "%lx" (fun x -> x)
+  let checksum_to = Llio.int32_to
+  let checksum_from = Llio.int32_from
+
+  let calculate s = Crc32c.calculate_crc32c s 0 (String.length s)
+  let update cs s = Crc32c.update_crc32c cs s 0 (String.length s)
+end
+
+module Md5 : ChecksumType = struct
```
NicolasT: This seems unused? If I'm not mistaken, the current code can't handle variations in the checksum format either. Consider a checksum type tag in the TLog format?

src/tools/checksum.ml
314eter: The None checksum is only necessary for compatibility with old tlogs. If they are printed with None and Some, the output of --dump-tlog will contain a lot of Some's (and never None), which seemed a bit superfluous to me.

src/tools/checksum.ml
314eter: Yes, if `t` is exposed as `Cs.t option`, which is better.

src/paxos/value.ml
NicolasT: Also related to some other comments: I think the API would be better suited if a checksum type was used as-is, wrapped into a non-opaque `option` whenever suitable, passing around values of the actual checksum type when `None` is not a valid value, using `option` on a per-function basis.

src/tools/checksum.ml
NicolasT: Agree (sort-of) in the `string_of` case. I don't really like this special-casing in `of_string` though. Is this function required anyway?

314eter: No, it's never used. I probably included it for the symmetry.

src/tlog/tlogcollection_test.ml
```diff
@@ -248,6 +249,29 @@ let test_validate_corrupt_1 (dn, tlf_dir, factory) =
...
+    )
+    (function
+      | Value.ValueCheckSumError _ -> Lwt.return ()
+      | exn ->
+        let () = ignore exn in
+        let msg = Printf.sprintf "it threw the wrong exception %s" "?" in
+        OUnit.assert_bool msg false;
+        Lwt.return ()
+    )
+  >>= fun() ->
```
NicolasT: `>>= fun () -> Lwt.return ()` is a no-op (actually it's the right-identity monad law: `m >>= fun v -> return v` equals `m`).

src/node/node_main.ml
314eter: The checksum is part of the printed value.

NicolasT: No, what I meant is: there are 2 kinds of checksums, the one for tlog entries (local checksums), and the rolling one. In order to avoid confusion, it might be useful to log which checksum the error is about exactly. 'Value has a checksum error' isn't very enlightening for a user.

PR #491, opened 2014-08-11
fix off by one error in copy_db_to_head

domsj:
Apparently copy_db_to_head keeps one more tlog than expected


toolslive:
OMG. (LGTM)

PR #477, opened 2014-07-07
undo "fix qshell extension for node ips list"

lienboe:

PR #493, opened 2014-08-11
1.6 for 1.7

domsj:


domsj:
updated with latest commits from 1.6

domsj:
note that the commits related to #477 result in an empty merge here...

domsj:
everything looks good on CI

NicolasT:
Seems like a (surprisingly!) small diff, which doesn't introduce anything fishy.

PR #497, opened 2014-08-14
fix universe.small_catchup.catchup.batched_with_failures failure

domsj:
see http://cimaster.incubaid.com/view/arakoon-git/job/arakoon-git-base/941/testReport/junit/universe.small_catchup/catchup/batched_with_failures/


NicolasT:
Ok if CI is happy :smile:

PR #500, opened 2014-08-18
update META & changelog for 1.7.4

domsj:
also reordered the changelog a bit to make it chronological

### DIFF
--- a/META
+++ b/META
@@ -1,7 +1,7 @@
 # -*- conf -*-
 
 description = "Arakoon client"
-version = "1.7.3"
+version = "1.7.4"
 exists_if = "arakoon_client.cma,arakoon_client.cmxa,arakoon_client.cmxs"
 requires = "threads lwt lwt.unix str oUnit quickcheck"
 archive(byte) = "arakoon_client.cma"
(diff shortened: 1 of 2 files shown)


NicolasT:
I'm a bit confused... Why does adding the new entry to the top of the changelog change so many things?

domsj:
A few entries in the changelog were not in chronological order, I 'rectified' that situation (as mentioned in my pull request comment above).
Do you want me to leave the ordering as it was?

NicolasT:
Woops, missed that remark. When doing these changes anyway, maybe get the spacing in sync as well? Some entries have 2 spaces after the name, before '<', some have only one. I don't know what the formatting rules say, but it's somewhat inconsistent now.

domsj:
You're right about those spaces, fixed that.
Do you want me to fix the extra space in your e-mail address too or is that somehow intentional?

NicolasT:
Hmh, never noticed that. Feel free to fix, my bad.

domsj:
No problem, adjusted that too

NicolasT:
Go

PR #502, opened 2014-08-19
Checksums

314eter:


doc/design/rolling-checksums.rst
```diff
@@ -0,0 +1,93 @@
+=================
+Rolling Checksums
+=================
+
+Problem
+=======
+If a node crashed, and failed to write some tlog entries to disk, this is not detected by Arakoon. The node announces it's in sync up to the last entry in the tlogs, even if other nodes diverged while the node was offline.
```
domsj: I think we should mention here that usually it should not be possible for this to happen. Usually means: fsync set to true (which is the default) and none of the layers below (file system, hardware) lie about fsync behaviour.

doc/design/rolling-checksums.rst
```diff
@@ -0,0 +1,93 @@
...
+Node1 and node2 crashed, and the last tlog entries were lost. They are restarted, while node0 is still offline. When node0 comes back, this will result in the following situation:
+
++--------------------------------------+----------------------------------+----------------------------------+
+| node0                                | node1                            | node 2                           |
++======================================+==================================+==================================+
+| 0:(Vm (node0,0.000000))              | 0:(Vm (node0,0.000000))          | 0:(Vm (node0,0.000000))          |
++--------------------------------------+----------------------------------+----------------------------------+
+| 1:(Vc ([Set;"a";1;"...";],false)     | 1:(Vc ([Set;"a";1;"...";],false) | 1:(Vc ([Set;"a";1;"...";],false) |
++--------------------------------------+----------------------------------+----------------------------------+
+| **2:(Vc ([Set;"b";1;"...";],false)** | **2:(Vm (node1,0.000000))**      | **2:(Vm (node1,0.000000))**      |
```
toolslive, line 36: So, node1 lost the 2 last tlog entries, and replaced them with others. This should not happen when you have an fsync between the writing of each entry, unless your mount options, file system, hardware are wanting. 

So is this whole set of changes some kind of runtime detection of bad configuration or a hardware lie-detector?

doc/design/rolling-checksums.rst
```diff
@@ -0,0 +1,93 @@
...
+Old value format
+----------------
+* 0xff (int32)
+* Value type (char 'c' or 'm')
+* Value details (depends on type)
+
+New value format
+----------------
+* 0x100 (int32)
+* Checksum (int32 if crc-32 is used)
```
toolslive, line 82: maybe you want to dabble into MICs and MACs.

src/msg/tcp_messaging.ml
```diff
@@ -69,7 +69,7 @@ class tcp_messaging
   my_addresses my_cookie (drop_it: drop_function)
   max_buffer_size ~stop =
 
-  let _MAGIC = 0xB0BAFE7L in
+  let _MAGIC = 0x53E7965CL in
```
domsj: I think we should keep the old magic in here as a comment

src/node/catchup.ml
```diff
@@ -114,8 +119,26 @@ let catchup_tlog (type s) ~tls_ctx ~stop other_configs ~cluster_id  mr_name ((mo
...
+    let _validate = !r_validate in
+    begin
+      r_validate := false;
+      _validate
+    end
+  in
+
+  let f_entry (i, value) =
+    let validate = validate () in
+    tlog_coll # log_value_explicit i value ~validate:validate false None >>= fun _ ->
```
domsj: You can write this as `tlog_coll # log_value_explicit i value ~validate false None >>= fun () ->`
note the `~validate` and `>>= fun ()`

src/node/catchup.ml
```diff
@@ -114,8 +119,26 @@ let catchup_tlog (type s) ~tls_ctx ~stop other_configs ~cluster_id  mr_name ((mo
   let mr_addresses = Node_cfg.client_addresses mr_cfg
   and mr_name = Node_cfg.node_name mr_cfg in
   Logger.info_f_ "getting last_entries from %s" mr_name >>= fun () ->
-  let head_saved_cb hfn =
-    Logger.info_f_ "head_saved_cb %s" hfn >>= fun () ->
+
+  let r_validate = ref false in
```
domsj: Shouldn't this start with `ref true` ?

src/node/catchup.ml
```diff
@@ -297,9 +320,28 @@ let verify_n_catchup_store (type s) ~stop me ?(apply_last_tlog_value=false) ((mo
...
+      end
     | i, Some j when i > j ->
-      catchup_store ~stop me ((module S),store,tlog_coll) too_far_i
+      let entry = ref None in
+      let check e = Lwt.return (entry := Some e) in
+      tlog_coll # iterate j (Sn.succ j) check >>= fun () ->
+      let store_cs = S.get_checksum store in
+      let tlog_cs =
+        match !entry with
+          | None -> None
```
domsj: this case should never happen, right?
so you may want to throw an exception

src/node/collapser_test.ml
```diff
@@ -63,7 +69,7 @@ let test_collapse_until (dn, tlf_dir, head_dir) =
   tlc # close () >>= fun () ->
   Lwt_unix.sleep 5.0 >>= fun () -> (* give it time to generate the .tlc *)
   (* now collapse first file into a tc *)
-  let storename = "head.db" in
+  let storename = Filename.concat test_dn "head.db" in
```
domsj, line 72: I'm a bit baffled ... how did it ever work before?
How did you notice this?

src/node/store.ml
```diff
@@ -882,6 +901,8 @@ struct
     >>= fun () ->
     _insert_updates store updates' kt >>= fun (urs:update_result list) ->
     _with_transaction store kt (fun tx -> _incr_i store tx) >>= fun () ->
+    let cs = Value.checksum_of value in
+    _with_transaction store kt (fun tx -> _set_checksum store cs tx) >>= fun () ->
```
domsj: I know you're already changing this, but as a reminder to myself: set_checksum should happen in the same transaction as the one that bumps the i just above

src/node/collapser_test.ml
314eter, line 72: I did not, git did...

src/paxos/value.ml
```diff
@@ -18,90 +18,155 @@ limitations under the License.
...
+let content_to buf = function
+  | Vc (us,synced) -> begin
+      Llio.char_to buf 'c';
+      Llio.bool_to buf synced;
+      Llio.list_to buf Update.to_buffer us
+    end
+  | Vm (m,l) -> begin
+      Llio.char_to buf 'm';
+      Llio.string_to buf m;
+      Llio.int64_to buf (Int64.of_float l)
```
domsj, line 38: I'm wondering if this can't lead to any problems in combination with https://github.com/Incubaid/arakoon/pull/502/files#diff-8312f0aa9ee0fb85222835bef6a0c9dcR164 where the lease_start is updated (and thus changes the serialization of the value) ?
Actually writing the lease start to disk here is (I think) the wrong thing to do.
I think this should ideally not be serialized or a dummy value - which should always be the same - should be serialized.

src/msg/tcp_messaging.ml
toolslive: the new magic has no magic. :(

src/paxos/value.ml
314eter, line 38: In `clear_self_master_set` and `fill_other_master_set`, values are changed without updating the checksum. This is ugly behaviour, but necessary because these checksums are stored in the store, and should validate with the original value (which is already saved in the tlog), not the changed one.

src/system/startup.ml
```diff
@@ -428,6 +424,72 @@ let interrupted_election () =
   Lwt_list.iter_s (_dump_tlc ~tlcs)   [node2; node3]
 
 
+let disk_failure () =
+  let lease_period = 2 in
+  let node0 = "disk_failure_slave0" in
+  let node1 = "disk_failure_slave1" in
+  let node2 = "disk_failure_was_master" in
```
domsj: How would a disk failure be involved (as these node names suggest)?
(I had power failure scenarios in mind)

src/system/startup.ml
```diff
@@ -428,6 +424,72 @@ let interrupted_election () =
...
+  let t_node1 = _make_run ~stores ~tlcs ~now ~get_cfgs ~values:[v0;v1] node1 () in
+  let run_previous_master = _make_run_rc ~stores ~tlcs ~now ~get_cfgs ~values:[v0;v1;v2;v3] node2 in
+  Logger.debug_ "start of scenario" >>= fun () ->
+  Lwt.ignore_result t_node0;
+  Lwt.ignore_result t_node1;
+  (* sleep a bit so the previous 2 slaves can make progress *)
+  Lwt_unix.sleep ((float lease_period) *. 1.5) >>= fun () ->
+  let t_previous_master = run_previous_master () >>= function
+    | 51 -> Lwt.return ()
+    | 0 -> Lwt.return (OUnit.assert_bool "the node should fail" false)
```
domsj: In case the node doesn't fail it will probably result in the test hanging, and not in executing this assertion.
Better would be to run the previous_master thread in a `Lwt_unix.with_timeout`

src/node/catchup.ml
314eter: Yes.

domsj: I would expect this sort of thing to be caught by a test .. is there a test we're missing?

src/system/startup.ml
```diff
@@ -428,6 +424,72 @@ let interrupted_election () =
...
+  let get_cfgs () = cluster_cfg in
+  let v0 = Value.create_master_value_nocheck ~lease_start:0. node2 in
+  let v1 = Value.create_client_value_nocheck [Update.Set("a","a")] false in
+  let v2 = Value.create_client_value_nocheck [Update.Set("b","b")] false in
+  let v3 = Value.create_client_value_nocheck [Update.Set("c","c")] false in
+  let tlcs = Hashtbl.create 5 in
+  let stores = Hashtbl.create 5 in
+  let now = Unix.gettimeofday () in
+
+  let t_node0 = _make_run ~stores ~tlcs ~now ~get_cfgs ~values:[v0;v1] node0 () in
```
domsj: Hmm, this uses the memtlogcollection ; I would have a safer warmer feeling if this test would be done with the regular tlogcollection.
@toolslive @NicolasT any opinions?

NicolasT: +1

src/paxos/value.ml
```diff
@@ -18,90 +18,155 @@ limitations under the License.
...
+    tlog_coll i
+    ?(lease_start = Unix.gettimeofday ())
+    m =
+  let c = Vm (m, lease_start) in
+  create_value tlog_coll i c
 
-let is_master_set  = function
-  | Vc _ -> false
-  | Vm _ ->  true
+let validate tlog_coll i (cso, c) =
```
domsj: I think `is_valid` would be a better name for this

src/tlog/tlc2.ml
```diff
@@ -794,12 +836,26 @@ class tlc2
...
+        Llio.input_int32 ic >>= fun crc ->
+        Llio.input_string ic >>= fun cmd ->
+        let value = Value.value_from (Llio.make_buffer cmd 0) in
+        if validate && not (Value.validate self i value)
+        then Lwt.fail (Value.ValueCheckSumError (i, value))
+        else
+          Sn.output_sn oc i >>= fun () ->
+          Llio.output_int32 oc crc >>= fun () ->
+          Llio.output_string oc cmd >>= fun () ->
+          let length = Int64.sub length (Lwt_io.position ic) in
```
domsj: Are we sure Lwt_io.position really gives the desired value here?
I think you want to get the position from this stream already on line 845 and use that too in your calculation...
(btw, is this code path under test?)

domsj:
Should we have some test for the upgrade scenario (as discussed earlier) ?
Ideally a cluster is started with all 1.7 nodes which are then 1 by 1 upgraded to 1.8.
After each node upgrade it should be possible to perform (e.g.) a test_and_set operation.

src/node/catchup.ml
314eter: The tests use `mem_tlogcollection` currently.

src/tlog/tlc2.ml
314eter: Moreover, this code assumes tlogs are uncompressed.

buildInSandbox.sh
```diff
@@ -2,14 +2,14 @@
...
-cp arakoon.native /opt/qbase3/apps/arakoon/bin/arakoon
-cp -r pylabs/extensions/* /opt/qbase3/lib/pymonkey/extensions/
-cp pylabs/Compat.py /opt/qbase3/lib/pymonkey/extensions/Compat.py
-
-mkdir -p /opt/qbase3/var/tests/arakoon_system_tests/
-touch /opt/qbase3/var/tests/arakoon_system_tests/__init__.py
-cp -r pylabs/test/* /opt/qbase3/var/tests/arakoon_system_tests/
-mkdir -p /opt/qbase3/lib/python/site-packages/arakoon/
-cp src/client/python/*.py /opt/qbase3/lib/python/site-packages/arakoon/
+sudo mkdir -p /opt/qbase3/apps/arakoon/bin/
```
NicolasT: These shouldn't be required.

PR #507, opened 2014-08-26
remove unused 'quiesced' field from Multi_paxos.constants

domsj:


domsj:
Discovered due to @lienboe asking me what this field was about...

PR #513, opened 2014-09-29
OVS-1325 - Added support for creating rpm packages

khenderick:
Added support for creating rpm packages

PR #514, opened 2014-09-29
OVS-1325 - Added support for creating rpm packages

khenderick:
Added support for creating rpm packages.


domsj:
As a historical note: this replaces #513 .

LGTM. As long as .deb build scripts are in here I think it's fine for .rpm build scripts to be here too.

NicolasT:
IMHO it lacks quite a couple of things:
- Packages tend to carry a `LICENSE` or `COPYING` file, which is installed in `/usr/share/doc/<packagename>`, using `%doc COPYING` in the `%files` section
- The `Release` field should most likely carry a `%{?dist}` suffix (`1%{?dist}`)
- The package description lacks build- & runtime-dependencies (`BuildRequires` and `Requires` fields)
- Missing `Source0` field (currently the package sources seem to come out of thin air)
- Consider setting `ExclusiveArch`, since Arakoon is basically only tested & supported on x86_64
- Use e.g. `%{_bindir}` instead of hard-coded `/usr/bin`
- Use `${buildroot}` instead of `$RPM_BUILD_ROOT`
- Consider adding a `%check` section which invokes the Arakoon unit tests (which ensures the build actually runs)
- Some other fields are missing:
  - `URL`
  - `%changelog`

A good read on RPM packaging: https://fedoraproject.org/wiki/How_to_create_an_RPM_package

NicolasT:
LGTM (assuming it works :wink: )

PR #515, opened 2014-10-02
make arakoon compile with ocaml 4.02 (for now with unsafe_string compile...

domsj:
...r flag)

Do note it no longer compiles with ocaml_version < (4,02)

I've also updated the CI scripts, but they don't work yet as quickcheck is not available in opam for ocaml 4.02.
I opened a new pr for Anil: https://github.com/avsm/ocaml-quickcheck/pull/4 ; after that I will make it available in opam.
If it takes too long I'll just add the updated quickcheck to our opam packages repository...

I ran the base tests locally and they succeeded.


domsj:
I'll close this PR and open a new one for 1.8 once #517 has been merged.

PR #521, opened 2014-10-06
fix client_lib + add --user-function to command line

domsj:


src/main/main.ml
```diff
@@ -292,6 +293,9 @@ let main () =
     ("--prefix", Arg.Tuple[set_laction PREFIX;
                            Arg.Set_string key;
                           ], "<prefix>: all starting with <prefix>");
+    ("--user-function", Arg.Tuple[set_laction USER_FUNCTION;
+                                  Arg.Set_string key;
+                                  Arg.Set_string value], "<name> <value>: invoke user function 'name' with arg 'value'");
```
toolslive: might be a bit tricky to do arbitrary binary params via the command line.

toolslive:
ok.

NicolasT:
Consider changing how `--user-function` works, by not taking an argument to the action, but reading `stdin` and use that as argument?

domsj:
Right now it is indeed not suited for arbitrary binary params.

I've changed passing in of the value to
`arakoon --user-function bla` to pass None to the user function
to which you can optionally add `-value foo` to pass in `Some "foo"`.

If we should later have the need for passing in binary params like this we could add another command line flag e.g. `-value-from-stdin` or `-value-from-file <filename>`.

PR #522, opened 2014-10-07
improve llio + minor update to client lib

domsj:


Makefile
```diff
@@ -38,6 +38,8 @@ install_client:
 	  _build/src/client/arakoon_client.cmi \
 	  _build/src/client/arakoon_remote_client.mli \
 	  _build/src/client/arakoon_remote_client.cmi \
+          _build/src/client/common.cmi \
+          _build/src/lib/std.cmi \
```
toolslive, line 42: That means client side dependency on test_utils and all the gifts it bares.

domsj, line 42: Would it make sense to split the arakoon_client lib into a part that's really for client and another part for plugins ?

toolslive:
k. fine by me ;)

PR #528, opened 2014-12-18
Llio.tupleX

domsj:


src/tools/llio.ml
```diff
@@ -323,6 +323,27 @@ let pair_from a_from b_from buf =
   let b = b_from buf in
   (a, b)
 
+let tuple3_from a_from b_from c_from buf =
```
toolslive, line 326: what about ?

``` OCaml
val threesum : (buf -> 'a) -> 
                        (buf -> 'b) -> 
                        (buf -> 'c) -> 
                        ('a -> 'b -> 'c -> 'd) -> 
                        'd
```

toolslive:
fine.

src/tools/llio.ml
domsj, line 326: Looks nice. Didn't need this right now, but I might add something similar in the future...

PR #530, opened 2015-01-02
Llio big endian

domsj:


toolslive:
we might want to keep travis alive too.

domsj:
Yes, I'll fix travis and merge once that works again

PR #531, opened 2015-01-13
log failure before exit

toolslive:
something like this 

```
Jan 13 09:27:13 0924: (main|fatal): nsm_host_plugin: Dynlink.Error "error loading shared library: /tmp/arakoon/arakoon_0/nsm_host_plugin.cmxs: undefined symbol: camlSlice"
```


domsj:
Ok

PR #534, opened 2015-01-20
Assert range empty

domsj:


domsj:
Fwiw, the failure on jenkins is unrelated.
(Some shaky test failing with Failure("No Master"))

toolslive:
lgtm

PR #551, opened 2015-05-20
More tlx

toolslive:
don't iterate over the first tlog after the head.db as it can be fetched in its entirety.


domsj:
LGTM if jenkins agrees

toolslive:
going to take another route.

PR #552, opened 2015-05-21
minor improvement to catchup_only

toolslive:


src/node/node_main.ml
```diff
@@ -198,7 +198,8 @@ let only_catchup (type s) (module S : Store.STORE with type t = s) ~tls_ctx ~nam
        let compressor = me.compressor in
        make_tlog_coll ~compressor
                       me.tlog_dir me.tlx_dir me.head_dir
-                      ~fsync:me.fsync name ~fsync_tlog_dir:me._fsync_tlog_dir >>= fun tlc ->
+                      ~fsync:false name ~fsync_tlog_dir:false
```
domsj, line 201: This change isn't needed given the changes in the tlogcollection, right?

domsj:
Ok if CI agrees :)

src/node/node_main.ml
toolslive, line 201: Indeed: that change is not needed, because the values are ignored in this code path anyway.

PR #559, opened 2015-06-12
Change behaviour of range like queries during transactions

domsj:
to fix a bug that could leave part of a transaction applied iso rolling back
- bugfix for the range_assertion


toolslive:
always a good sign if you can throw away things. 
fine.

PR #566, opened 2016-02-21
can changes go back to incubaid?

despiegk:
or do I make a mistake here?


domsj:
I think you can do as you please ... the incubaid repo is all yours.
Do note that the license has changed from apache 2 to agpl.
And you're merging our 1.9 branch into a 1.8 branch on this repo.

